PR #2, opened 2017-08-16
Bg fix detached head

pngowda:
@sergej-koscejev  thanks for the review comments. I have made the required changes as per the comments.


sergej-koscejev:
Thanks! I've fixed the formatting before merging.

PR #10, opened 2017-12-11
Add explicit types in POM helper

coolya:
Encountered problems in build scripts that the method couldn't get called
I added explicit types here now. Not sure if the "internal" Configuration
is valid to use. But seems to be the only way to get access to resolved
dependencies.


sergej-koscejev:
> Not sure if the "internal" Configuration is valid to use. But seems to be the only way to get access to resolved dependencies.
Why? The Configuration API contains `getResolvedConfiguration()` method. https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/Configuration.html

coolya:
Thanks for the pointer I overlooked it in the interface. Replaced it.

PR #11, opened 2017-12-13
Avoid failed distro build due to posix file permissions when running on windows

wsafonov:


sergej-koscejev:
Why are you calling CreateDmg task on Windows? It is a Mac-only task.

sergej-koscejev:
Declined after discussion.

PR #13, opened 2018-01-22
add task to download mps in the browser

coolya:
Sometimes its handy to allow the user to simply download the current
MPS version. The code checks the current operating system and then tries
to open the download uri in the browser.


coolya:
I basically encountered two problems during the last week:

- We have some documentation in a customers confluence that stats which MPS version to use and where to download it. It frequently becomes out of date because people forget to update it. Especially when it's just a minor version bump with no migrations. 

- I had to setup a repo for a customer workshop and I wanted to make the "on boarding" as smooth as possible. By simply telling them "run this gradle command and it will take care of everything".

If it is of little use for others I can just leave it in the customer project.

sergej-koscejev:
Okay, I think it's fine to merge.

PR #20, opened 2018-06-22
generate user project

coolya:
Migrated our build scripts to kotlin gradle scripts.

Added a project plugin and classes that run a headless MPS form the command line to generate a MPS project without requiring MPS build scripts. 

The plugin uses a two stage approach to do this. The first stage is a normal gradle project plugin without any dependencies to MPS. 

The second stage is a java command line application that is lunched by the gradle plugin. The command line application requires MPS to be on the classpath. Therefore the first stage will setup the classpath correctly and then lunch a second JVM via a `JavaExc` task. 

A minimal build script to generate a MPS project with no external plugins would look like this: 

```
apply plugin: 'generate-models'

configurations {
    mps
}

ext.mpsVersion = '2017.3.5'

generate {
    projectLocation = new File("./mps-prj")
    mpsConfig = configurations.mps
}

dependencies {
    mps "com.jetbrains:mps:$mpsVersion"
}
```

The plugin will behind the scenes create the required tasks to copy MPS to a location where it is started from and also takes care of faking the build number correctly because for some reason in a headless application the branding.jar is not loaded and the ApplicationInfo is not initialised. 

Todos: 

- [x] add description to the `generate` task and add it to a group 
- [x] add documentation to the readme 
- [x] think about declaring output files for the generate task


execute-generators/src/main/kotlin/de/itemis/mps/gradle/generate/Helper.kt
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+}
+
+private fun createScript(proj: Project, models: List<org.jetbrains.mps.openapi.model.SModel>): IScript {
+
+    val allUsedLanguagesAR: AsyncResult<List<SLanguage>> = AsyncResult()
+
+    proj.modelAccess.runReadAction {
+        allUsedLanguagesAR.setDone(models.map { it.module }.distinct().flatMap { it.usedLanguages }.distinct())
```
sergej-koscejev: We should also take into account the languages that the used languages extend

src/main/kotlin/de/itemis/mps/gradle/generate/Plugin.kt
```diff
@@ -0,0 +1,92 @@
...
+
+            val genConfig = configurations.maybeCreate("genRuntime")
+            genConfig.isVisible = false
+
+            afterEvaluate {
+                val mpsVersion = extension
+                        .mpsConfig
+                        .resolvedConfiguration
+                        .firstLevelModuleDependencies.find { it.moduleGroup == "com.jetbrains" && it.moduleName == "mps" }
+                        ?.moduleVersion ?: throw GradleException("MPS configuration doesn't contain MPS")
```
sergej-koscejev: What if I package and use my RCP instead of MPS here?

src/main/kotlin/de/itemis/mps/gradle/generate/Plugin.kt
```diff
@@ -0,0 +1,92 @@
...
+
+open class GenerateMpsProjectPlugin : Plugin<Project> {
+
+    override fun apply(project: Project) {
+        project.run {
+
+            val extension = extensions.create("generate", GeneratePluginExtensions::class.java)
+            val mpsLocation = extension.mpsLocation ?: File(project.buildDir, "mps")
+
+            val genConfig = configurations.maybeCreate("genRuntime")
```
sergej-koscejev: Have you considered using a detached (anonymous) configuration instead of a named one? https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/ConfigurationContainer.html#detachedConfiguration-org.gradle.api.artifacts.Dependency...-

src/main/kotlin/de/itemis/mps/gradle/generate/Tasks.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.gradle.api.tasks.InputDirectory
+import org.gradle.api.tasks.InputFile
+import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.TaskAction
+import java.io.File
+import java.io.PrintWriter
+import java.util.*
+
+
+open class FakeBuildNumberTask: DefaultTask() {
```
sergej-koscejev: What is this used for?

execute-generators/src/main/kotlin/de/itemis/mps/gradle/generate/Helper.kt
coolya: If not sure but you might be correct. It's pretty hard to track what MPS is doing in the `ModulesClusterizer` and friends. But seems like `ModulesCluster.fillEdges` is in fact doing a ` new SLanguageHierarchy(myLanguageRegistry, moduleUsedLanguages).getExtended()`. Will change the implementation.

src/main/kotlin/de/itemis/mps/gradle/generate/Plugin.kt
coolya: I wasn't sure here which one to use. The named one might have the advantage that the user could use it in a dependency resolution strategy block and force a specific version. While the detached one is nice because it would conflicts with anything the user has created.

src/main/kotlin/de/itemis/mps/gradle/generate/Tasks.kt
coolya: The problem is that for some reason the `ApplicationInfo` isn't initialised properly. That causes `PluginManagerCore.BUILD_NUMBER` to be null. 

In this case the `PluginManagerCore` resorts to `BuildNumber.currentVersion()` which finally calls into `BuildNumber.fromFile()`. That allows us to place a `build.txt` in the root of the home path (see `PathManager.getHomePath()`). Which is then used to load the build number from. 

I just saw that at least for 2018.2 a newer version of the platform is shipped which allows to get a similar behaviour via setting `idea.plugins.compatible.build` property.

src/main/kotlin/de/itemis/mps/gradle/generate/Plugin.kt
coolya: Se my comment below.

src/main/kotlin/de/itemis/mps/gradle/generate/Tasks.kt
sergej-koscejev: Can you add this information as a doc comment to the class?

src/main/kotlin/de/itemis/mps/gradle/generate/Plugin.kt
sergej-koscejev: I would use a detached config and let the user specify the dependency, similar to how JDK is selected in the CreateDmg task.

coolya: Done.

src/main/kotlin/de/itemis/mps/gradle/generate/Tasks.kt
coolya: Added it to the `Plugin.kt` file where the task is used.

execute-generators/src/main/kotlin/de/itemis/mps/gradle/generate/Helper.kt
coolya: Done.

README.md
```diff
@@ -122,3 +122,50 @@ present). `destination` is then generated based on the properties.
...
+}
+
+dependencies {
+    mps "com.jetbrains:mps:$mpsVersion"
+}
+```
+
+Parameters:
+* `mpsConfig` - the configuration used to resolve MPS. Currently only vanilla MPS is supported and no custom RCPs.
+  Custom plugins are supported via the `pluginLocation` plugin.
```
sergej-koscejev: `pluginLocation` plugin -> `pluginLocation` property

README.md
```diff
@@ -122,3 +122,50 @@ present). `destination` is then generated based on the properties.
...
+* `mpsLocation` - optional location where to place the MPS files.
+* `plugins` - optional list of plugins to load before generation is attempted.
+  The notation is `new Plugin("someID", "somePath")`. Where the first parameter is the plugin id and the second the `short (folder) name`.
```
sergej-koscejev: future work: maybe use simple strings with ':' as the separator for plugins (like Gradle uses for dependencies), this would make the syntax easier.  If the id and the folder are the same, you could even just mention one of them.

coolya:
@sergej-koscejev can you re approve it? The repo was configured to dismiss review state when new pushes to PR are made and I fixed the readme file.

Issue #22, opened 2018-07-13
Add migration checker task


coolya:
A task that checks if any modules in the project requires migration.

Issue #21, opened 2018-07-13
Add model checking task


coolya:
A requirement that often pops up from users is the ability to run the model checker from the commandline. 

For instance to fail the build and further generation of a project if it contains any model checking errors.

rafaroca:
I want to do exactly that, fail the build when there are model checker errors. I'd highly appreciate this feature

rafaroca:
@coolya are you actively working on this?

coolya:
@rafaroca yes we are. Currently reviewing some infrastructure changes in #20 that we will need in this task as well. #20 should be merged by the end of this week. Then adding this feature should be straight forward.

rafaroca:
@coolya I see that #20 is now merged. Are you planning to work on the model checker task?

rafaroca:
@coolya any progress on this issue? are you working on this, is there another task blocking? thanks

coolya:
@rafaroca we have a working version of it wich is not yet open source. I expect it to be published in the next two week from now.

PR #25, opened 2018-07-30
RFC: Infrastructure to execute code with a project

coolya:
This PR is not meant to be merged yet. I opened it to get feedback as early as possible and to discuss potential further work.

# Infrastructure to execute code with a project

As we are more and more often required to execute certain tasks during the build the  `IdeaEnvironment` becomes the go to solution. We also have some further plans to add new functionality to this plugin that requires access to a loaded MPS plugin (#21, #22). This PR aims to provide a common infrastructure to make these functionality as easy to use as possible. To do so the common bits of the `execute-generators` subproject were extracted.

## Java API 

The Java API is a single function `executeWithProject` it has been written in Kotlin but is perfectly callable from normal java code as a static function but is more idiomatic Kotlin and java. It might feel strange to java programmers to work with such API especially as it's more a functional style than normal OO. 

The function takes all arguments required to setup the `IdeaEnvironment` plus a function that gets the project once it's loaded. The function can produce a value which is returned to the caller. The `executeWithProject` function also takes care of the complete lifecycle of the project and the `IdeaEnvironment`. 

### Discussion 

- Is the idiomatic Kollin API *good enough* for java developers or do we need a more OO like facet? 
- Are the use cases were caching the the `IdeaEnvironment` is essential over several project opening? 
- Error logging/handling is currently mostly missing

## Command line API

In addition to the actual Java API I tried to *standardise* a command line API. As most of our current use cases spawn a new JVM rather then loading all of MPS into the same JVM as gradle we often need to pass information around via command line arguments. To avoid reinventing the wheel and having multiple incompatible implementations floating around I created a extensible command line argument definition. It takes all required parameters and allows passing them to a overload of the `executeWithProject` directly. 

Application specific arguments can be added by extending the `Args` class. 

### Discussion 
- Do we want to standardise this interface? 
- The reuse is limited to Kotlin as the argument parser isn't really usable from oder JVM languages. 

## Reuse

The `project-loader` is reusable beyond the scope of this project. We could point others who want to do something with a headless MPS to this plugin and it's command line interface. That's why I would suggest publishing the `project-loader` as a separate maven publication. This comes with the downside that we will have to be more carefully of introducing breaking changes as we will not have all of the users under our control.

Open tasks: 
- [ ] update `GenerateMpsProjectPlugin` to use new command line API.


coolya:
I would appreciate any feedback on this PR.

sergej-koscejev:
I'll review this on Monday.

project-loader/src/main/kotlin/Args.kt
```diff
@@ -0,0 +1,41 @@
...
+ * subclasses to [executeWithProject] is directly supported.
+ */
+open class Args(parser: ArgParser) {
+
+    val plugins by parser.adding("--plugin",
+            help = "plugin to to load. The format is --plugin=<path>:<id>")
+    { toPlugin(this) }
+
+    val macros by parser.adding("--macros",
+            help = "macro to define. The format is --macro=<name>:<value>")
```
sergej-koscejev: so is this `--macros` or `--macro`?

project-loader/src/main/kotlin/Loader.kt
```diff
@@ -0,0 +1,138 @@
...
+ * @param pluginLocation optional location where plugins lo load are located. This is for additional plugins, the plugins
+ * located in the pre-installed pluging location (usual the "plugins" folder of MPS) are still considered.
+ * @param plugins optional list of plugins to load. Path is the what MPS calls the `short (folder) name` in it's build
+ * language. The id is the plugin id defined in the plugin descriptor. Both are required because MPS supports multiple
+ * plugins the same location.
+ * @param macros optional list of path macros to define before the project is opened
+ * @param buildNumber optional build number that is used to determine if the plugins are compatible. Only guaranteed to
+ * work with MPS 2018.2+ but might work in earlier versions as well.
+ */
+fun <T> executeWithProject(project: File,
```
sergej-koscejev: How about bundling the environment options into a class? I'm on the fence about this myself but maybe the class will be a natural place to add methods to in the future.

project-loader/src/main/kotlin/Loader.kt
```diff
@@ -0,0 +1,138 @@
...
+                           plugins: List<Plugin> = emptyList(),
+                           macros: List<Macro> = emptyList(),
+                           pluginLocation: File? = null,
+                           buildNumber: String? = null,
+                           action: (Project) -> T): T {
+
+    if (pluginLocation != null) {
+        logger.info("overriding plugin location with: ${pluginLocation.absolutePath}")
+        System.setProperty(PROPERTY_PLUGINS_PATH, pluginLocation.absolutePath)
```
sergej-koscejev: Should we set the property to the old value when we're finished? To clean up after ourselves.

project-loader/src/main/kotlin/Loader.kt
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    if (buildNumber != null) {
+        logger.info("setting build number to \"$buildNumber\"")
+        System.setProperty(PROPERTY_PLUGINS_COMPATIBLE_BUILD, buildNumber)
```
sergej-koscejev: Ditto.

project-loader/src/main/kotlin/Loader.kt
sergej-koscejev: Ah ok, I see below that the Args class is playing that role. It's probably enough then.

sergej-koscejev:
I would like to have this functionality accessible from Maven so I'm looking into this now.

A few years ago I wrote [mps-maven-plugin](https://github.com/JetBrains/mps-maven-plugin) where I used a similar approach but I used the GeneratorWorker and Script classes from mps-tool.jar to start the generator.

This would allow us to save some code and it seems to be a more "official" entry point since GeneratorWorker/Script are used by the `generate` Ant task. What do you think about that approach?

coolya:
@sergej-koscejev I think the major reason was that the MPS-generate task does require you to give it a calculated make script with all the libraries and also to calculate the chunks manually. Which the `LangaugeWorkbenchMakeService` does for you.

coolya:
@sergej-koscejev Sorry for the very late reply. If I recall the details correctly, I think the major reason was that the MPS-generate task does require you to give it a calculated make script with all the libraries and also to calculate the chunks manually. Which the `LangaugeWorkbenchMakeService` does for you.

coolya:
@sergej-koscejev Sorry for the very late reply. If I recall the details correctly, I think the major reason was that the MPS-generate task does require you to give it a calculated make script with all the libraries and also to calculate the chunks manually. Which the `LangaugeWorkbenchMakeService` does for you.

coolya:
@sergej-koscejev Sorry for the very late reply. If I recall the details correctly, I think the major reason was that the MPS-generate task does require you to give it a calculated make script with all the libraries and also to calculate the chunks manually. Which the `LangaugeWorkbenchMakeService` does for you.

coolya:
@sergej-koscejev Sorry for the very late reply. If I recall the details correctly, I think the major reason was that the MPS-generate task does require you to give it a calculated make script with all the libraries and also to calculate the chunks manually. Which the `LangaugeWorkbenchMakeService` does for you.

coolya:
While manually testing the build scripts I found a bug in the argument parser and fixed it. I wanted to setup automated tests for the gradle scripts but this turned out to be harder than expected. 

I needed to update to gradle 4.10 to get support for snapshot plugin versions as a start. 

The next problem is testing against the plugin version that was just build. I tried as suggested in the documentation by using a composite build in the test project that includes the plugin build, but I was unable to get gradle to find the plugin with this setup. 

The only workaround I found is that I publish a snapshot version of the plugin to maven local and the use maven local as a plugin repository in the tests. This seems to work but will required us to do some changes in the CI to have the tests execute correctly. I thought about using gradles testkit to implement the tests but then getting the required mps projects into the build location wasn't really straight forward. So I opted for adding independent gradle projects to the repo that use maven local to get the plugin. 

To test this manually you need to do the following: 

- In the root project execute `./gradlew publishToMavenLocal publishGenerate-modelsPluginMarkerMavenPublicationToMavenLocal` this will put the gradle plugin (which is mostly a pom file) and the other artefacts into maven local
- in `test/generate-*` run `./gradlew generate` this will pickup the plugin from maven local eventually succeed generating the project. 

I added two smoke tests, one that generates a simple java class and a second one that generate a build.xml file from the mps build project.

coolya:
Form my perspective this is ready to be merged. See my above comments to see what I changed from the last review. I would also suggest that we now bump the version number to 1.1 which we forgot to do last time.

If anyone used the command line generation with the current version these changes would break the command line interface to the spawned java subprocess. So just to be sure I bumped the version to 1.1.

PR #28, opened 2018-09-28
macos builds will succeed even if same-named volume exists

abstraktor:
This commit allows the plugin to build on macos regardless of what volumes are already mounted.

# Background

So far, I needed to eagerly close the current distro if I want to build a new one, because otherwise it will not succeed. The reason is that the mpsdmg.sh and mpsdmg.pl assume the temporary mount point to have the same name as the final one, which is not the case if there is a name collision. Then the mount point will be sth like `/Volumes/SecurityAnalyst 1` in our case; instead of `/Volumes/SecurityAnalyst`.

# A sidenote on BKGD_alias

The assembled .DS_Store contains the correct alias to the node of the provided background file, yet the original path seems to be persisted as well (as a hexdump of it shows). If the temporary mount point during assembly differs, this name will end up in that alias. This is the path that we provide in line 18 of the perl script. 

This is still the correct value because HFS aliases function by reference. Since aliases are stable on renaming, this should not be an issue, and I tested it to work regardless of the eventual mount name. I am not sure though, why the original path is stored in an alias at all.

# Steps to reproduce

Given I build an RCP on macos
And I mount the dmg file
When I rebuild the RCP
And the dmg file is still mounted
Then the build will fail

```
…
Build Name: SecurityAnalyst.app
Vol Name: SecurityAnalyst
512+0 records in
512+0 records out
524288 bytes transferred in 0.005408 secs (96945874 bytes/sec)
Creating unpacked r/w disk image SecurityAnalyst in /var/folders/44/71zyf5hj0p1_l5gxmgwt8q6w0000gp/T/tmp.QHdpwPpN.dmg...
Mounting unpacked r/w disk image...
Mounted as /dev/disk3.
Updating disk image styles...
rm: /Volumes/SecurityAnalyst/DSStorePlaceHolder: No such file or directory


FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':code:packageMoraADDistribution_mac'.
> Process 'command '/var/folders/44/71zyf5hj0p1_l5gxmgwt8q6w0000gp/T/groovy-generated-6889508253266095071-tmpdir/mpsdmg.sh'' finished with non-zero exit value 1

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 5m 39s
12 actionable tasks: 9 executed, 3 up-to-date
```


sergej-koscejev:
I have no way to test it so I'm just going to trust @abstraktor that it works.

sergej-koscejev:
@abstraktor please merge this

PR #29, opened 2018-10-22
git versioning: add method to check if git is available

coolya:
Added a helper method to check if git is available so that build scripts can handle this case gracefully instead of catching the exceptions thrown form the underlying API.


coolya:
`git rev-list` behaves different in git 1.x. It's pretty unlikely that we will every encounter such ancient git versions but still wanted to make sure.

coolya:
I double checked with the windows version of git and `startsWith("git version 2")` should be fine.

Issue #30, opened 2018-11-04
Report model checking results as tests


coolya:
Currently error reporting for the model checker is pretty limited. Getting the exact model where the check failed is only possible by parsing the log files. It would be much better if we could integrate the model checker with normal test infrastructure like junit to report the failures through junit xml files. 

This would allow CI to show the exact models where a check has failed. 

One possible implementation could be done with parameterised tests. For instance [kotlintest](https://github.com/kotlintest/kotlintest) `FunSpec` or `Describe Spec` could be used to executed a test for each model that is model checked. 

An open questions is how we could get the required information from the gradle plugin into the actual test task. As far as I can see most test frameworks depend on a runner application that hosts the tests and is responsible for test discovery etc. Most probably the current approach by passing the required information via the command line will then no longer work. 

Alternatively we could write our own junit xml file that contains the test result and is then used by CI and/or gradle.

coolya:
see #48 as an implementation of this.

PR #31, opened 2018-11-04
model checking support

coolya:
Implements #21 and is based on #25 hence #25 needs to be merged first. 

This pull request adds support to run the model checker directly from gradle on a project or subset of models from a project. It's configuration is closely aligned with the generator from command line part that we already have in the project.

In general the gradle part is not much different, a simple plugin with extensions for configuration and the a command line _backend_ that does the real work. Some of the decisions why to use certain classes are documented in the code directly but my main goal was to stay away from anything the uses the `ModelCheckerSetting` from MPS to avoid any interferences with the settings of the user.

Currently there is no way to configure which checks are executed and how the errors are reported. Everything is sent to log4j. In theory users could use the `log.xml` file in the MPS installation to customise the logging. Users could also use log4j [configuration](https://logging.apache.org/log4j/2.x/manual/configuration.html) via properties of env variables to redirect the log out put somewhere else. 

Customisation can be done by changing properties of the `checkmodels` task which is a [JavaExec](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/JavaExec.html) task.

## Future Work 

- Integrate with testing infrastructure (#30) 
- Better usage documentation with advanced examples


arimer:
Is there a way how I can run the modelcheck test? After deploying the modelchecker locally with `modelcheck:publishPluginMavenPublicationToMavenLocal` I'll get an artifact with the version: `modelcheck-2018.2.5.2-SNAPSHOT`.

However, the build.gradle file for the **/test/modelcheck** project tries to use a modelcheck plugin with the following version:
` id("modelcheck") version "1.0-SNAPSHOT"` which it obviously can't find under the local repos.

Am I doing something wrong, or do we need to adapt the required migrationcheck version in the **/test/modelcheck** project?

README.md
```diff
@@ -180,3 +180,74 @@ Parameters:
...
+* `plugins` - optional list of plugins to load before generation is attempted.
+  The notation is `new Plugin("someID", "somePath")`. Where the first parameter is the plugin id and the second the `short (folder) name`.
+* `pluginLocation` - location where to load the plugins from. Structure needs to be a flat folder structure similar to the
+  `plugins` directory inside of the MPS installation.
+* `models` - optional list of models to generate. If omitted all models in the project will be generated. Only full name
+  matched are supported and no RegEx or partial name matching.
+* `macros` - optional list of path macros. The notation is `new Macro("name", "value")`.
+* `projectLocation` - location of the MPS project to generate.
+* `errorNoFail` - report errors but do not fail the build.
+* `warningsAsError` - handles warnings as errors and will fail the build if any is found when `errorNoFail` is not set.
```
sergej-koscejev: There's an inconsistency between `--warning-as-error` and `warningsAsError` setting (the latter has an 's').

test/generate-build-solution/build.gradle.kts
```diff
@@ -0,0 +1,38 @@
+group = "test.de.itemis.mps.gradle.generate"
+version = "1.0-SNAPSHOT"
+
+plugins {
+    id("generate-models") version "1.0-SNAPSHOT"
```
sergej-koscejev: This should now be 1.1-SNAPSHOT. @arimer is this maybe the problem that was preventing you from testing the plugin?

test/generate-simple/build.gradle.kts
```diff
@@ -0,0 +1,37 @@
+group = "test.de.itemis.mps.gradle.generate"
+version = "1.0-SNAPSHOT"
+
+plugins {
+    id("generate-models") version "1.0-SNAPSHOT"
```
sergej-koscejev: Should be 1.1-SNAPSHOT as well

test/modelcheck/build.gradle.kts
```diff
@@ -0,0 +1,30 @@
+import java.net.URI
+
+group = "test.de.itemis.mps.gradle.modelcheck"
+version = "1.0-SNAPSHOT"
+
+
+plugins {
+    id("modelcheck") version "1.0-SNAPSHOT"
```
sergej-koscejev: And here too (1.1-SNAPSHOT)

PR #32, opened 2019-01-10
gradle 5

coolya:
I updated our build to use gradle 5.

As part of the update I noticed that we are loading two versions of kotlin at runtime. Kotin 1.2 through the intellij platform and 1.3 as a direct dependency. I refactored that build scripts to use a consistent version of kotlin through the whole project. 

I also enabled that "warnings as errors" for the kotlin build that we fail the build in these cases.

In addition I replaced the deprecated `AsyncResult` with `Promise` and made a kotlin switch exhaustive to get rid of the warnings.

In addition I incremented the minor version number to prevent clients from accidentally getting the new version.


settings.gradle.kts
```diff
@@ -1,3 +1,5 @@
 rootProject.name = "mps-gradle-plugin"
 include("execute-generators")
-include("project-loader")
\ No newline at end of file
+include("project-loader")
+
+enableFeaturePreview("STABLE_PUBLISHING")
```
arimer: I think we can remove this line. The spec clains we do not need them any longer: 
see: https://docs.gradle.org/current/userguide/upgrading_version_4.html#changes_5.0

Issue #34, opened 2019-02-12
Allow project wide script classpath configufation


coolya:
Right now all tasks that extend the `RunAntScript` task need to specify the classpath with `scriptClasspath` which typically includes the ant_lib as well as the path to the `tools.jar` file. 

This is duplicated all over the place when these tasks are declared with the same value. We should have a way to define the default classpath and arguments for all `RunAntScript` tasks. These defaults are used if now specific values are specified for a task.

arimer:
Do you have already a concrete idea how you will solve it? I'd be interested.

coolya:
PR incoming for this.

PR #35, opened 2019-02-13
allow to specify defaults for classpath and args

coolya:
Introduced two new properties to control the projectwide defaults for
the classpath and the arguments passed to the JVM that executes the
`RunAntScript´ task. See updated README for more details.

Added `RunAntScript` tasks documentation to the readme.

Implements #34

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ buildscript {
     }
 
     dependencies {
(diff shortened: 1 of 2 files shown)


arimer:
This change works like a charm in our research project.

PR #36, opened 2019-02-14
update to MPS 2018.3

coolya:
Updated the dependency on MPS to MPS 2018.3.3 from MPS 2018.2.6.
This updadate required us to use kotlin 1.3 because the intellij
platform migrated to that kotlin version. Kont 1.3 brings a stricter
null checker so we had to adapt some code in the helpers that interact
with the intellij platform.

Removed the deprecated `getInstance` call for the `FacetRegistry` with a
proper `Project.getComponent` call.


abstraktor:
- pulled from master
- set MPS version to 2018.3.6 and the CI is fine with it.

What's left to be done for this to be merged? (SecA is on 2018.3.6 and thus needs this to use the modelchecker)

test/modelcheck/mps-prj/.mps/modules.xml
```diff
@@ -2,7 +2,7 @@
 <project version="4">
   <component name="MPSProject">
     <projectModules>
-      <modulePath path="$PROJECT_DIR$/solutions/my.solution/my.solution.msd" folder="" />
+      <modulePath path="$PROJECT_DIR$/$mypath$/my.solution/my.solution.msd" folder="" />
```
coolya: Please don't do this. Using Macros in `modules.xml` is not supported in MPS 2019.1+.

abstraktor: How about undoing this in the test project and just adding this to the readme example then (so that others won't need to search for it again, as I had to)?

abstraktor:
How is it determined then, for which mpsVersion the master build will run? (I'll need 2018.3.6…)

execute-generators/src/main/kotlin/de/itemis/mps/gradle/generate/Helper.kt
```diff
@@ -56,15 +57,17 @@ private fun createScript(proj: Project, models: List<org.jetbrains.mps.openapi.m
     val scb = ScriptBuilder()
 
     scb.withFacetNames(allUsedLanguages
-            .mapNotNull { registry.getLanguage(it) }
-            .mapNotNull { it.getAspect(MakeAspectDescriptor::class.java) }
-            .flatMap { it.manifest.facets() }
-            .map { it.name }
+            ?.mapNotNull { registry.getLanguage(it) }
```
sergej-koscejev: I'd prefer a message here if allUsedLanguages turn out to be null.
Looks like the proposed version would silently do nothing in this case?

execute-generators/src/main/kotlin/de/itemis/mps/gradle/generate/Helper.kt
```diff
@@ -116,7 +119,7 @@ fun generateProject(parsed: GenerateArgs, project: Project): Boolean {
 
     val modelsToGenerate = ftr.get()
 
-    return makeModels(project, modelsToGenerate)
+    return if(modelsToGenerate != null) makeModels(project, modelsToGenerate) else true
```
sergej-koscejev: In what situation would modelsToGenerate be null? Should this maybe be reported to the user rather than silently returning `true`?

test/generate-build-solution/build.gradle.kts
```diff
@@ -15,7 +15,7 @@ repositories {
 
 val mps = configurations.create("mps")
 
-val mpsVersion = "2018.2.5"
+val mpsVersion = "2018.3.3"
```
sergej-koscejev: what about 2018.3.6 here?

test/generate-simple/build.gradle.kts
```diff
@@ -15,7 +15,7 @@ repositories {
 
 val mps = configurations.create("mps")
 
-val mpsVersion = "2018.2.5"
+val mpsVersion = "2018.3.3"
```
sergej-koscejev: What about 2018.3.6 here?

.gitignore
```diff
@@ -16,4 +16,6 @@ mps-prj/solutions/my.build/build.xml
 test/generate-build-solution/mps-prj/solutions/my.build/build.xml
 .DS_Store
 workspace.xml
+**/source_gen**
+**/classes_gen/**
```
sergej-koscejev: This seems inconsistent with the line above.
Maybe just ignore `**/*_gen*/`?

sergej-koscejev: Because there could also be `test_gen`

abstraktor: I'd suggest `**/*_gen/` then so that `my_fancy_file_with_gen_in_name.txt` is not included

sergej-koscejev: a file won't be included, only a directory. There's also `source_gen.caches/`

execute-generators/src/main/kotlin/de/itemis/mps/gradle/generate/Helper.kt
abstraktor: afaics it only happens when the [AsyncPromise](https://github.com/JetBrains/intellij-community/blob/master/platform/util/concurrency/org/jetbrains/concurrency/AsyncPromise.kt) gets explicitly canceled or its value is set to null. Which both doesn't happen. Let's log an error and return false here to fail the build. This way, if some of this reasoning was wrong, we'll find out.

.gitignore
abstraktor: good to know

execute-generators/src/main/kotlin/de/itemis/mps/gradle/generate/Helper.kt
abstraktor: Should resolve with https://github.com/mbeddr/mps-gradle-plugin/pull/36/commits/48c1c05207fec32b41c1d5ce75ebde1e99926f62

execute-generators/src/main/kotlin/de/itemis/mps/gradle/generate/Helper.kt
abstraktor: should resolve with https://github.com/mbeddr/mps-gradle-plugin/pull/36/commits/ffd675d293197310b6b6bfef315509514881f0ef

PR #44, opened 2019-07-05
don't attach branch prefix to version for builds on maintenance branches

wsafonov:
This allows to publish de.itemis.mps.(execute-generators|modelcheck|project-loader) on maintenance branches without branch prefix, so that these published artifacts can be resolved by simply using de.itemis.mps:modelcheck:$mpsVersion version string.


wsafonov:
@sergej-koscejev I've introduced a utility method in GitBasedVersioning for stripping maintenance branch info from the version - this way we can use it also in user projects where same functionality is required.

PR #48, opened 2019-09-05
modelcheck: add support for JUnit XML report

coolya:
The model check plugin can not report its results as a JUnit xml file
which seems the defacto standard for test reporting in CI. For each
model that is checked it will create a testcase entry and report if as
failed if any model checking error is reported in that model. If multiple
error are reported in the same model they will all get reported as a
failed assertion in that test case.

Reporting is controlled via the new `junitFile` configuration for the
plugin. If the configuration is present the grade plugin will report
into that file.

The classes representing a JUnit XML file were placed in the common
package of the project loader plugin for easier reuse if we want to 
report from other plugins results as a junit xml file. 

I also fixed an NPE in case a property wasn't defined before the plugin
would executed. In this case resetting the value to `null` would cause
an exception in `System.setProperty`.

### DIFF
--- a/modelcheck/src/main/kotlin/Main.kt
+++ b/modelcheck/src/main/kotlin/Main.kt
@@ -189,7 +189,7 @@ fun modelCheckProject(args: ModelCheckArgs, project: Project): Boolean {
         errorCollector.result.map { printResult(it, project, args) }
 
         if (args.xmlFile != null) {
-            writeJuniXml(project.projectModels, errorCollector.result, project, File(args.xmlFile!!))
+            writeJunitXml(project.projectModels, errorCollector.result, project, File(args.xmlFile!!))
         }
     }
 
--- a/modelcheck/src/main/kotlin/Main.kt
+++ b/modelcheck/src/main/kotlin/Main.kt
@@ -86,10 +86,14 @@ fun printResult(item: IssueKindReportItem, project: Project, args: ModelCheckArg
 }
 
 
-fun writeJunitXml(models: Iterable<SModel>, results: Iterable<IssueKindReportItem>, project: Project, file: File) {
+fun writeJunitXml(models: Iterable<SModel>,
+                  results: Iterable<IssueKindReportItem>,
+                  project: Project,
+                  warnAsErrors: Boolean,
+                  file: File) {
 
     val allErrors = results.filter {
-        it.severity == MessageStatus.ERROR
+        it.severity == MessageStatus.ERROR || (warnAsErrors && it.severity == MessageStatus.WARNING)
     }
     val errorsPerModel = allErrors
             .filter {
@@ -189,7 +193,7 @@ fun modelCheckProject(args: ModelCheckArgs, project: Project): Boolean {
         errorCollector.result.map { printResult(it, project, args) }
 
         if (args.xmlFile != null) {
-            writeJunitXml(project.projectModels, errorCollector.result, project, File(args.xmlFile!!))
+            writeJunitXml(project.projectModels, errorCollector.result, project, args.warningAsError, File(args.xmlFile!!))
         }
     }
 
--- a/modelcheck/src/main/kotlin/Main.kt
+++ b/modelcheck/src/main/kotlin/Main.kt
@@ -1,15 +1,16 @@
 package de.itemis.mps.gradle.modelcheck
 
+import com.fasterxml.jackson.dataformat.xml.XmlMapper
 import com.xenomachina.argparser.ArgParser
+import com.xenomachina.argparser.default
 import com.xenomachina.argparser.mainBody
+import de.itemis.mps.gradle.junit.*
 import de.itemis.mps.gradle.project.loader.Args
 import de.itemis.mps.gradle.project.loader.executeWithProject
 import jetbrains.mps.checkers.*
 import jetbrains.mps.errors.MessageStatus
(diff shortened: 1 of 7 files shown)


project-loader/src/main/kotlin/JUnit.kt
```diff
@@ -0,0 +1,105 @@
...
+        val classname: String,
+        @field:JacksonXmlProperty(isAttribute = true)
+        val assertions: Int? = null,
+        @field:JacksonXmlProperty(isAttribute = true)
+        val status: String? = null,
+        @field:JacksonXmlProperty(isAttribute = true)
+        val time: String? = null,
+        @field:JacksonXmlElementWrapper(useWrapping = false)
+        @field:JacksonXmlProperty(localName = "skipped")
+        @field:JsonInclude(JsonInclude.Include.NON_NULL)
```
coolya: I know it looks strange to apply a JSON annotation to a XML serialized class but the Jackson XML serializer will respect this annotation as well and omit the value in the XML file when it's null.

project-loader/src/main/kotlin/Loader.kt
```diff
@@ -123,7 +123,12 @@ fun <T> executeWithProject(project: File,
     
     // cleanup overridden property values to the state that they were before.
     propertyOverrides.forEach {
-        System.setProperty(it.first, it.second)
+
+        // if a property wasn't set before the value is "null"
+        // setting null as a value for a System property will result in a NPE
+        if(it.second != null) {
```
coolya: We could also set it to `""` (empty string)  I'm not 100% sure what is better. Any opinions?

project-loader/src/main/kotlin/JUnit.kt
```diff
@@ -0,0 +1,105 @@
+package de.itemis.mps.gradle.junit
```
coolya: Data structure was taken from here: https://llg.cubic.org/docs/junit/

project-loader/src/main/kotlin/JUnit.kt
```diff
@@ -0,0 +1,105 @@
...
+                     val tests: Int,
+                     @field:JacksonXmlProperty(isAttribute = true)
+                     val disabled: Int? = null,
+                     @field:JacksonXmlProperty(isAttribute = true)
+                     val errors: Int? = null,
+                     @field:JacksonXmlProperty(isAttribute = true)
+                     val failures: Int? = null,
+                     @field:JacksonXmlProperty(isAttribute = true)
+                     val id: Int,
+                     @field:JacksonXmlProperty(localName = "package", isAttribute = true)
```
coolya: `package` is a reserved keyword so we have to work around this.

coolya:
This implements #30

modelcheck/src/main/kotlin/Main.kt
```diff
@@ -80,7 +85,70 @@ fun printResult(item: IssueKindReportItem, project: Project, args: ModelCheckArg
     }
 }
 
-fun modelCheckProject(args: ModelCheckArgs, project: Project) : Boolean {
+
+fun writeJuniXml(models: Iterable<SModel>, results: Iterable<IssueKindReportItem>, project: Project, file: File) {
```
arimer: Small typo here. Should be writeJuni**t**Xml I guess.

coolya:
```suggestion
fun writeJunitXml(models: Iterable<SModel>, results: Iterable<IssueKindReportItem>, project: Project, file: File) {
```

Github has this awesome feature to suggest changes to the pr. Its super useful for typos :)

arimer: Sorry did not know about this! Will use it next time 👍

modelcheck/src/main/kotlin/Main.kt
```diff
@@ -119,6 +187,10 @@ fun modelCheckProject(args: ModelCheckArgs, project: Project) : Boolean {
 
         // We need read access here to resolve the node pointers in the report items
         errorCollector.result.map { printResult(it, project, args) }
+
+        if (args.xmlFile != null) {
+            writeJuniXml(project.projectModels, errorCollector.result, project, File(args.xmlFile!!))
```
arimer:
```suggestion
            writeJunitXml(project.projectModels, errorCollector.result, project, File(args.xmlFile!!))
```

coolya:
@arimer added the documentation

coolya:
added the clear properties call and noticed that we weren't handling `warningAsError` correctly in the JUnit export.

coolya:
One thing I just noticed when playing around with these changes in a project is that we fail the build in case of any model checking error is reported. Which is fine for the normal use case I think. But in case we are exporting the JUnit I think we don't want to do that. Also MPS doesn't fail the build in the cases but reports the results in xml file. What do you think shall gracefully exit the process in case we are writing the JUnit xml regardless of errors being present?

abstraktor:
If the build fails for a red test or model error, it should do this for both or none of the cases. In fact, I'd like to be able to turn on "fail build if tests or checker fail", when running a CLI build locally. But that's a different story.

coolya:
@sergej-koscejev @abstraktor what you want is already possible:

> `errorNoFail` - report errors but do not fail the build.

I was asking if `errorNoFail` should be the default when running with `junitFile` or if the user will have to specify it explicitly.

PR #50, opened 2019-09-17
Bugfix: add missing "generate" target

sergej-koscejev:
I assumed that it was a dependency of "assemble" so I deleted it but it
was a mistake.


sergej-koscejev:
@coolya @arimer Somebody please push the button here, I was too clever and tried to remove some targets from BuildLanguages and TestLanguags but it turns out they are required.

arimer:
Thanks for adding it back!

PR #52, opened 2019-09-19
Avoid CDATA in the JUnit output of modelchecker

sergej-koscejev:


coolya:
Any reason why we would like to avoid them? Afaik also MPS wraps these element data with a CData tag?

sergej-koscejev:
@coolya Model checker on a customer project produces messages with ]]> in them and Jackson complains when serializing them.

I mentioned

coolya:
I see, sorry I didn't look at the commit msg of that commit. 

I haven't tested what happens when the output from `stdout` contains xml tags. I would assume Jackson will then simply escape these correctly.

PR #53, opened 2019-12-04
modelcheck: add max heap size property

coolya:
In some scenarios the OS reported memory limit isn't the same as the one
applied to individual processes, e.g. when running inside a container.
To prevent the process from being killed setting the max heap value is
required.

### DIFF
--- a/src/main/kotlin/de/itemis/mps/gradle/modelcheck/Plugin.kt
+++ b/src/main/kotlin/de/itemis/mps/gradle/modelcheck/Plugin.kt
@@ -15,6 +15,7 @@ open class ModelCheckPluginExtensions: BasePluginExtensions() {
     var warningAsError = false
     var errorNoFail = false
     var junitFile: File? = null
+    var maxHeap: String? = null
 }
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/de/itemis/mps/gradle/modelcheck/Plugin.kt
```diff
@@ -15,6 +15,7 @@ open class ModelCheckPluginExtensions: BasePluginExtensions() {
     var warningAsError = false
     var errorNoFail = false
     var junitFile: File? = null
+    var maxHeap: String? = null
```
arimer: If I understood the syntax correctly, it means that heapsize is optional. Correct?

coolya: Thats correct. All existing project will work without specifying it. If you do not specify it the JVM will try to figure out the max heap size by its usual means. Looking into `JAVA_OPS` env variable for instance or checking the OS memory available.

PR #54, opened 2020-01-06
Update to MPS 2019.2

wsafonov:
MPS 2019.2 has minor changes in API that had to be reflected in the plugin code. Also, there is an issue with some src jars that caused an exeption during MPS startup - see commits and the code for further details.

### DIFF
--- a/execute-generators/build.gradle.kts
+++ b/execute-generators/build.gradle.kts
@@ -37,13 +37,15 @@ version = if (project.hasProperty("forceCI") || project.hasProperty("teamcity"))
 }
 
 
-val mpsConfiguration = configurations.create("mps")
-
 dependencies {
     implementation(kotlin("stdlib-jdk8", version = kotlinVersion))
     implementation("com.xenomachina:kotlin-argparser:$kotlinArgParserVersion")
-    mpsConfiguration("com.jetbrains:mps:$mpsVersion")
-    compileOnly(mpsConfiguration.resolve().map { zipTree(it) }.first().matching { include("lib/*.jar") })
+    compileOnly("com.jetbrains:mps-openapi:$mpsVersion")
+    compileOnly("com.jetbrains:mps-core:$mpsVersion")
+    compileOnly("com.jetbrains:mps-tool:$mpsVersion")
+    compileOnly("com.jetbrains:mps-messaging:$mpsVersion")
+    compileOnly("com.jetbrains:platform-api:$mpsVersion")
+    compileOnly("com.jetbrains:platform-concurrency:$mpsVersion")
     implementation(project(":project-loader"))
 }
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -23,7 +23,7 @@ val nexusUsername: String? by project
 val nexusPassword: String? by project
 
 val kotlinArgParserVersion by extra { "2.0.7" }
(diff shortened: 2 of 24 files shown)


execute-generators/build.gradle.kts
```diff
@@ -37,13 +37,15 @@ version = if (project.hasProperty("forceCI") || project.hasProperty("teamcity"))
 }
 
 
-val mpsConfiguration = configurations.create("mps")
-
 dependencies {
     implementation(kotlin("stdlib-jdk8", version = kotlinVersion))
     implementation("com.xenomachina:kotlin-argparser:$kotlinArgParserVersion")
-    mpsConfiguration("com.jetbrains:mps:$mpsVersion")
-    compileOnly(mpsConfiguration.resolve().map { zipTree(it) }.first().matching { include("lib/*.jar") })
```
arimer: Why don't we need to include the lib/*.jars any longer?

coolya:
Looks good to me. The source jar indexing bug is really odd and the workaround quite ugly but I think the only way how we can work around this. Hope fully with preindexed files coming later this year they won't screw it up again. 

Thanks for replacing the big mps dependency with the individual jar files from the nexus which makes is much fast to work with the code base in the IDE!

execute-generators/build.gradle.kts
wsafonov: Those Jars from the complete mps distro are replaced by smaller Maven publications based on the MPS distro, which drastically reduces the size of all dependencies needed to be downloaded for the plugin build.

wsafonov:
@coolya I have tested the plugin with 2019.3 - the issue with broken src jars seemed to be resolved there. Basically we can leave that last commit with a workaround in 2019.2 version only - I would prepare another PR for 2019.3 based on the previous commit and this full PR could be then merged into 2019.2 maintenance branch.

@sergej-koscejev not quite sure why do we need that - in my undestanding the plugin would remain compatible with 2019.1 after this change. That is, if some one would use the latest 1.2+ version of the gradle plugin providing mpsConfig for 2019.1, then simply an old version of the plugin runtime library (modelcheck or execute-generators) would be resolved and used in the runtime. It would be only an issue if later on there will be a change in the plugin making it incompatible with an older runtime lib - someone would need to go and create an update for that old version then.

wsafonov:
@sergej-koscejev I have to correct myself in the previous comment - actually you're right that the new plugin would not be compatible with the previous MPS version, because of the classpath changes (locations of checkmodels and httpsupport plugins jars changed). So the current plugin version built for MPS 2019.1 would not be compatible with MPS 2019.2 anyway. But we could change the classpath computation in the new plugin to make it backward compatible with MPS 2019.1, so everyone using "1.2+" version would get the new plugin for MPS 2019.2, which would still support 2019.1. For that we would need to add all jars from `plugins` directory to the classpath. That would blow up the classpath length causing well-known troubles under windows. But apart from that I don't see specific reasons why that should not work. The start-up performance might be slightly affected, but not sure if that would be really significant. In any case that would be a trade-off for getting backward compatibility with MPS.

sergej-koscejev:
@wsafonov and I decided to increase the version number and then later increase the version number again when we upgrade to 2019.3 and the ugly workaround becomes unnecessary and is removed.

PR #55, opened 2020-02-04
WIP Refactor/mps20193

jenzb:
Updated plugin for MPS 2019.3 and fix some issues that occurred with gradle 6.

### DIFF
--- a/src/main/groovy/de/itemis/mps/gradle/GenerateLibrariesXml.groovy
+++ b/src/main/groovy/de/itemis/mps/gradle/GenerateLibrariesXml.groovy
@@ -4,6 +4,7 @@ import groovy.xml.MarkupBuilder
 import org.gradle.api.DefaultTask
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputFile
+import org.gradle.api.tasks.Optional
 import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.TaskAction
 
@@ -11,7 +12,7 @@ class GenerateLibrariesXml extends DefaultTask {
     @InputFile
     File defaults
 
-    @InputFile
+    @InputFile @Optional
     File overrides
 
     @OutputFile
--- a/src/main/groovy/de/itemis/mps/gradle/GenerateLibrariesXml.groovy
+++ b/src/main/groovy/de/itemis/mps/gradle/GenerateLibrariesXml.groovy
@@ -2,6 +2,7 @@ package de.itemis.mps.gradle
 
 import groovy.xml.MarkupBuilder
 import org.gradle.api.DefaultTask
+import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputFile
 import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.TaskAction
@@ -10,6 +11,7 @@ class GenerateLibrariesXml extends DefaultTask {
     @InputFile
     File defaults
 
+    @InputFile
     File overrides
 
     @OutputFile
--- a/src/main/kotlin/de/itemis/mps/gradle/Common.kt
+++ b/src/main/kotlin/de/itemis/mps/gradle/Common.kt
@@ -3,6 +3,10 @@ package de.itemis.mps.gradle
 import org.gradle.api.GradleException
 import org.gradle.api.Project
 import org.gradle.api.artifacts.Configuration
+import org.gradle.api.file.FileCollection
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputFiles
+import org.gradle.api.tasks.Optional
 import java.io.File
 
 data class Plugin(
@@ -23,6 +27,7 @@ open class BasePluginExtensions {
     var macros: List<Macro> = emptyList()
     var projectLocation: File? = null
     var debug = false
+    var executable: Any? = null
 }
 
 fun argsFromBaseExtension(extensions: BasePluginExtensions): MutableList<String> {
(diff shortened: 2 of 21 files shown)


wsafonov:
Can you create a branch directly in the mps-gradle-plugin repo? Makes it easies to review and also the build pipeline for the PR would work...

exe222:
> Can you create a branch directly in the mps-gradle-plugin repo? Makes it easies to review and also the build pipeline for the PR would work...

I wanted to, but @coolya asked me to follow the "official" way. I have no permission to work directly on the mps-gradle-plugin repo, yet.

src/main/kotlin/de/itemis/mps/gradle/Common.kt
```diff
@@ -23,6 +27,7 @@ open class BasePluginExtensions {
     var macros: List<Macro> = emptyList()
     var projectLocation: File? = null
     var debug = false
+    var executable: Any? = null
```
coolya: Why is this `Any`? What is this used for? 

If its a `File` pointing to some executable then it should be a `File` if its path as a string then `String`.

src/main/kotlin/de/itemis/mps/gradle/modelcheck/Plugin.kt
```diff
@@ -91,6 +91,10 @@ open class ModelcheckMpsProjectPlugin : Plugin<Project> {
                 tasks.create("checkmodels", JavaExec::class.java) {
                     dependsOn(resolveMps)
                     args(args)
+                    if (extension.executable != null) {
```
coolya: Why do you want to use a different java executable than the one currently on `JAVA_HOME`?

src/main/groovy/de/itemis/mps/gradle/GenerateLibrariesXml.groovy
```diff
@@ -10,6 +11,7 @@ class GenerateLibrariesXml extends DefaultTask {
     @InputFile
     File defaults
 
+    @InputFile
```
coolya: We can't set this to `InputFile` because it must no be `null` in hat case otherwise gradle will complain hat an input is null. It definitely also needs to be `Optional`.

src/main/kotlin/de/itemis/mps/gradle/Common.kt
exe222: The variable name, idea and type copy pasted from RunAntScript task. I just wanted to be consistent...

src/main/kotlin/de/itemis/mps/gradle/modelcheck/Plugin.kt
exe222: As discussed briefly with @sergej-koscejev, instead of messing with the environment variable we wanted to force e.g. Java11 from the  client side. (See also MPS slack conversation)

Background: We have migrated mps-gradle-plugin to 2019.3 and want to use the model check in our project with this plug-in.
CI build of our project runs fine with Java 11.
But the model check still seems to use Java 8 and fails:
Exception in thread "main" java.lang.UnsupportedClassVersionError: jetbrains/mps/errors/item/IssueKindReportItem has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
If I manually set the env.JAVA_HOME parameter to Java11, then we get quite far and the check runs through. I thought, that it should work without messing with environment variables. just like our client project.

But if just setting JAVA_HOME to Java11 is the way to go, we could do this as well.

src/main/groovy/de/itemis/mps/gradle/GenerateLibrariesXml.groovy
exe222: Okay, will change accordingly. I was not sure about the proper annotations but gradle6 was complaining about them. :)

src/main/kotlin/de/itemis/mps/gradle/modelcheck/Plugin.kt
```diff
@@ -91,6 +91,10 @@ open class ModelcheckMpsProjectPlugin : Plugin<Project> {
...
+                        executable(extension.executable!!)
+                    }
+
```
coolya: I see the point. I think doing this change in general is fine. As the MPS execution for the task requires a jvm11 anyway we should also check in case there is no path to the jvm supplied that we are actually running in a jvm11 at the moment. 
```suggestion
                    if (extension.executable != null) {
                          executable(extension.executable!!)
                    } else {
                        if (JavaVersion.current() != JavaVersion.VERSION_11) logger.error("MPS requires Java 11 but current JVM uses ${JavaVersion.current()}, starting MPS will most probably fail!")
                    }
```

Since you changed the `Plugin` data class we should also apply this change to the [Plugin.kt](https://github.com/mbeddr/mps-gradle-plugin/blob/master/src/main/kotlin/de/itemis/mps/gradle/generate/Plugin.kt) of the generate plugin.

exe222:
Implemented according to comments. Builds OK locally. 
The actual model check in our client project is not working, though. This may be due to missing dependencies on our side.

wsafonov:
> Implemented according to comments. Builds OK locally.
> The actual model check in our client project is not working, though. This may be due to missing dependencies on our side.

Running the modelcheck on the customer project would be interesting before merging. Let's have a look at the problem you have together.

wsafonov:
The changes look good to me. There is only one more thing regarding the current plugin version. Due to the change in 2019.3 that requires Git/SVN plugins to be on the classpath to be loaded (as opposed to 2019.2 where the plugin loader could load those plugins directly), the previous 1.3 version of the plugin becomes incompatible with MPS 2019.3. Hence I would like to increase the plugin version to 1.4 in this PR. @sergej-koscejev agree with that?

sergej-koscejev:
> Hence I would like to increase the plugin version to 1.4 in this PR. @sergej-koscejev agree with that?

Yes.

PR #61, opened 2020-04-09
make overrides files optional

### LINKED ISSUE #60: GenerateLibrariesXml broken in 1.4
The GenerateLibrariesXml throws an exception in 1.4 if the overrides files does not exist.
It used to work in 1.2

```
org.gradle.api.tasks.TaskValidationException: A problem was found with the configuration of task ':generateLibrariesXml'.
        at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.report(ValidatingTaskExecuter.java:72)
        at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.execute(ValidatingTaskExecuter.java:58)
        at org.gradle.api.internal.tasks.execution.SkipEmptySourceFilesTaskExecuter.execute(SkipEmptySourceFilesTaskExecuter.java:108)
        at org.gradle.api.internal.tasks.execution.ResolveBeforeExecutionOutputsTaskExecuter.execute(ResolveBeforeExecutionOutputsTaskExecuter.java:67)
        at org.gradle.api.internal.tasks.execution.ResolveAfterPreviousExecutionStateTaskExecuter.execute(ResolveAfterPreviousExecutionStateTaskExecuter.java:46)
        at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:94)
Caused by: org.gradle.api.InvalidUserDataException: File '/Users/xxxxx/projectlibraries.overrides.properties' specified for property 'overrides' does not exist.
        at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.report(ValidatingTaskExecuter.java:69)
        ... 34 more
```

coolya:
Seems like we compiled 1.3+ of this plugin unintentionally against a newer version of the gradle api which caused the `@InputFile` annotation to process more strict. Affected by that was the `GenerateLibrariesXml` task which would now fail when the overrides file didn't exist. 

The annotation wasn't required anyway because the input dependency was anyway added in the setter of the `overrides` property. 

Added dependency locks to all projects so that dependency updates are no longer implicit but explicit and a new build doesn't unintentionally use a new version of a dependency. 

Added github actions as a secondary build where we would against all supported java versions.

Added tests for the `GenerateLibrariesXml` task.

Closes #60


build.gradle.kts
```diff
@@ -90,6 +100,31 @@ publishing {
     }
 }
 
+subprojects {
+    dependencyLocking {
+        lockAllConfigurations()
+    }
+}
+
+tasks.register("createClasspathManifest") {
```
sergej-koscejev, line 109: What is this task for? I found it in [Gradle docs](https://docs.gradle.org/current/userguide/test_kit.html) saying it's needed for very old versions of Gradle (before 2.13).

coolya: I wasn't able to get automatic class path working. Might be a problem with the way how the root build project is configured.

PR #70, opened 2020-07-16
Refactor CreateDmg and BundleMacosJdk: use Property<..>, migrate to Kotlin

### LINKED ISSUE #69: Avoid eager resolving artifact in CreateDmg#jdkDependency
https://github.com/mbeddr/mps-gradle-plugin/blob/a4a8f3cf7204799e3cd9dd2e331c8f3692f65c0e/src/main/groovy/de/itemis/mps/gradle/CreateDmg.groovy#L55-L60

Gradle tasks should not perform artifact resolution at the configuration time as the task might even be non-required at all.

vlsi:
fixes #69 

The PR includes the following changes:
* Use Gradle's `Property` API so the tasks support lazily generated inputs. For instance, jdk dependency is downloaded only in case the task is actually executed
* Migrate to Kotlin
* `File.createTempDir` is replaced with `build` directory to better conform with Gradle conventions
* README cleanup: added Kotlin DSL samples, fixed that JDK dependency requires .zip rather than .tgz
* Added `cleanTemporaryDirectories` option for debugging purposes (if set to `false` it would skip removing the temporary directories used for preparing the output artifact)
* Added integration test to verify CreateDmg and BundleMacosJdk tasks


vlsi:
The createDmg test passes locally, however it fails in GitHub Actions :(

```
arch: posix_spawnp: perl5.18: Bad CPU type in executable
```

arimer:
Hello Vladimir. Thanks for contributing! 🎉 
Do you mind adding a short discription to this PR to just sumup your idea. This is mostly useful for documentation purposes and might help others during code review.

vlsi:
@arimer , well, I thought It was summarized in the PR title. I duplicated it to the description as well.

src/main/kotlin/de/itemis/mps/gradle/CreateDmg.kt
```diff
@@ -0,0 +1,85 @@
...
+        objects: ObjectFactory,
+        providers: ProviderFactory,
+        layout: ProjectLayout
+) : BaseBundleMacOsTask(objects, providers, layout) {
+    @InputFile
+    val backgroundImage = objects.fileProperty()
+
+    @OutputFile
+    val dmgFile = objects.fileProperty()
+            .convention(layout.buildDirectory.file("createDmg/$name/$name.dmg"))
```
arimer, line 24: Shouldn’t the dmg file  not be placed under the dmg dir (L42) or do I missunderstand the usage of `dmgDir`?

src/main/kotlin/de/itemis/mps/gradle/CreateDmg.kt
```diff
@@ -0,0 +1,85 @@
...
+            throw GradleException("Value of dmgFile must end with .dmg but was $dmgFile")
+        }
+
+        val scripts = arrayOf("mpssign.sh", "mpsdmg.sh", "mpsdmg.pl",
+                "Mac/Finder/DSStore/BuddyAllocator.pm", "Mac/Finder/DSStore.pm")
+        val signingInfo = arrayOf(signKeyChainPassword, signKeyChain, signIdentity)
+        val scriptsDir = scriptsDir.get().asFile
+        val dmgDir = dmgDir.get().asFile
+        try {
+            rmdirs(dmgDir)
```
arimer, line 58: Just out of curisosity, we remove the `dmgDir`here  just to have cleanup before creating?

src/test/kotlin/de/itemis/mps/gradle/test/CreateDmgTest.kt
```diff
@@ -0,0 +1,161 @@
+package de.itemis.mps.gradle.test
```
arimer, line 1: I really like the idea having a minimal RCP test which can be executed. 
Could you maybe just drop some lines in the readme about the test idea and the setup? 
This would be great and could help others in the future to update/enrich the test if required.

arimer (changes requested):
Thanks a lot for this contribution I have still some open questions and points which are commented on the code.

src/main/kotlin/de/itemis/mps/gradle/BundleMacosJdk.kt
```diff
@@ -0,0 +1,50 @@
...
+        val scriptsDir = scriptsDir.get().asFile
+        val tmpDir = tmpDir.get().asFile
+        try {
+            val scriptName = "bundle_macos_jdk.sh"
+            BundledScripts.extractScriptsToDir(scriptsDir, scriptName)
+            scriptsDir.listFiles()?.let {
+                for (file in it) {
+                    println(file.absolutePath)
+                }
+            }
```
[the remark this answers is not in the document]
vlsi: That was indeed a debugging output

src/main/kotlin/de/itemis/mps/gradle/CreateDmg.kt
```diff
@@ -0,0 +1,85 @@
...
+            project.exec {
+                executable = "./mpssign.sh"
+                workingDir = scriptsDir
+                args("-r", rcpArtifact.get().asFile, "-o", dmgDir, "-j", jdk.get().asFile)
+
+                if (signingInfo.all { it.isPresent }) {
+                    args("-p", signKeyChainPassword.get())
+                    args("-k", signKeyChain.get().asFile)
+                    args("-i", signIdentity.get())
+                } else if (signingInfo.any { it.isPresent }) {
```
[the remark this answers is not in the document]
vlsi: The condition here is right.
The logic is
```kotlin
if (signingInfo.all { it.isPresent }) {
...
} else if (signingInfo.any { it.isPresent }) { // <- this
   there are SOME fields set and SOME missing => raise error
}
```

src/main/kotlin/de/itemis/mps/gradle/CreateDmg.kt
```diff
@@ -0,0 +1,85 @@
...
+            BundledScripts.extractScriptsToDir(scriptsDir, *scripts)
+            project.exec {
+                executable = "./mpssign.sh"
+                workingDir = scriptsDir
+                args("-r", rcpArtifact.get().asFile, "-o", dmgDir, "-j", jdk.get().asFile)
+
+                if (signingInfo.all { it.isPresent }) {
+                    args("-p", signKeyChainPassword.get())
+                    args("-k", signKeyChain.get().asFile)
+                    args("-i", signIdentity.get())
```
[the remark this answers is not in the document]
vlsi: These statements are additional arguments.

src/main/kotlin/de/itemis/mps/gradle/CreateDmg.kt
vlsi: `dmgDir` existed previously, and it is a temporary directory for preparing the dmg contents.

`dmgFile` is an output `dmg` file which can be placed whenever you want.
Now I think it might make sense to use `buildDirectory.file("distributions/createDmg/$name/$name.dmg")` for the default name of the output file.

src/main/kotlin/de/itemis/mps/gradle/CreateDmg.kt
vlsi: Is here to avoid stale content to affect the task.

build.gradle.kts
```diff
@@ -1,5 +1,6 @@
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 import java.net.URI
+import org.gradle.api.tasks.testing.logging.TestExceptionFormat
```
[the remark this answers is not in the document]
arimer: Maybe @coolya or @sergej-koscejev have an other oppinion.

src/test/kotlin/de/itemis/mps/gradle/test/CreateDmgTest.kt
vlsi: I'm puzzled here. Can you please clarify what is not clear?

I'm using a standard approach to use Gradle plugins: org.gradle.testkit.runner.GradleRunner
The test creates Gradle project within a temporary directory and executes a single task there.

build.gradle.kts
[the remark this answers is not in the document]
vlsi: Frankly speaking, the version is puzzling here, so I need your input here.
Of course, I can bump minor version to 5

src/main/kotlin/de/itemis/mps/gradle/CreateDmg.kt
vlsi: However, it might be better to integrate with https://docs.gradle.org/current/userguide/distribution_plugin.html somehow.

sergej-koscejev (commented):
I can't give a good review on this because I don't use these tasks, they run on our build server to produce a DMG for mbeddr RCP and either they work properly or nobody seems to care about DMGs. I also don't understand if there's a particular problem that this PR is supposed to solve.

@vlsi do you intend to maintain these tasks in the future? Have you had problems with these tasks?

Otherwise I'm tending against merging this PR because I'd like to avoid fixing what isn't broken.

sergej-koscejev (changes requested):
Sorry, I overlooked that this fixes #69. In that case let's merge this, but please bump version to 1.5 as @arimer mentioned.

vlsi:
@sergej-koscejev , sorry for the delay. I've rebased the PR and I bumped the version.

Do you think it can be merged?

sergej-koscejev:
I can't merge it because the macOS builds fail. @vlsi any idea why?

vlsi:
> macOS builds fail

I think the build fails since macOS scripts are outdated (see https://github.com/mbeddr/mps-gradle-plugin/issues/99)
However, I was not sure if that is GitHub's issue or if the scripts went out of date indeed.

vlsi:
So I guess the ways to proceed are:
* temporarily skip macOS from build matrix (or mark it to ignore errors)
* update macOS scripts

WDYT?

sergej-koscejev:
I'm fine with updating the macOS scripts. @coolya do you have any opinion?

coolya:
> I'm fine with updating the macOS scripts. @coolya do you have any opinion?

I'm fine with updating the files.

vlsi:
Well, I did check the scripts from JetBrains, and it looks like they are worse than the scripts in mps-gradle-plugin :((
They use positional arguments instead of named ones, they hard-code certain assumptions (e.g. JetBrains,  .png suffix, and so on).

As far as I understand, there's no point in building `.dmg` without bundling JDK. Java evolves at extreme pace nowadays, so it looks like the only sane way to release MPS-based IDEs (and `.dmg` in particular) is to bundle the workable JRE right into the distribution.

`generic.zip` might be helpful, however, it looks like "always bundle JRE" is the way to go for macOS, Windows, Linux. 

So I suggest:
1. update mps-gradle-plugins to accommodate macOS notarization
2. heal somehow the Perl failure (e.g. borrow the relevant bits from JetBrains Platform)
3. drop `bundleJdk` task altogether and let `createDmg` do everything (create the dmg with jre inside and sign it). An alternative option is to split `createDmg` (basically avoid shell scripts) into smaller tasks (perform "create dmg", "sign", "notarize" in different Gradle tasks), however, it would probably require more effort, and the gains are not obvious.


WDYT?

sergej-koscejev:
Closing as apparently abandoned.

Issue #62, opened 2020-04-21
Consider deprecating and removing GenerateLibrariesXml task


sergej-koscejev:
The task is used to generate a libraries.xml file with libraries pointing to custom locations, with the possibility to override some of those locations locally.

I think that instead people should simply check libraries.xml into version control. Library paths can contain project-relative paths so library `mylib` can be added with path `$PROJECT_DIR$/build/dependencies/mylib` for example, and Gradle can be configured to put the library there.

For the override functionality, one can instead open a different version of `mylib` in MPS before opening the main project. The override functionality allows one to change the project library path locally to point to `../mylib-development` but the added value over just opening it in MPS first seems to be very small.

vlsi:
@sergej-koscejev , how about the following:

1) User adds a Gradle configuration for the libraries
2) User adds dependencies to the configuration

```
val mpsPlugins by configurations.creating { ... }

dependencies {
    mpsPlugins("de.itemis.mps:extensions:$mpsExtensionsVersion")
    mpsPlugins("com.mbeddr:platform:$mbeddrBuildNumber")
}
```

3) Then `GenerateLibrariesXml` resolves the configuration, copies the libraries to `build/..` and generates (or even updates) libraries.xml with the proper location of the libraries.

I guess it would help to keep the library declarations in a single place.

On top of that, `mps-gradle-plugin` could declare a default configuration for `MPS libraries`. For instance `mpsLibraries`.
Then user could just add the dependencies, and `generateXmlLibraries` would update `.mps/libraries.xml`

WDYT?

sergej-koscejev:
@vlsi I think that generating this file doesn't save that much work, in my experience the set of used libraries doesn't change as much. If it does change often for you then you can add a parent directory such as `build/dependencies` as a project library and have multiple libraries inside that directory, MPS will find them. Adding each library separately is not really necessary.

vlsi:
> I think that generating this file doesn't save that much work

Well, the dependencies still have to be received and unpacked to a folder, don't they?
So it looks natural that the plugin should both download the dependencies, and generate the relevant `libraries.xml`

sergej-koscejev:
> > I think that generating this file doesn't save that much work
> 
> Well, the dependencies still have to be received and unpacked to a folder, don't they?
> So it looks natural that the plugin should both download the dependencies, and generate the relevant `libraries.xml`

[...]

PR #73, opened 2020-07-18
Migration to MPS 2020.1

wsafonov:
This PR migrates mps-gradle-plugin to MPS 2020.1.3.

There were no changes related to the MPS start-up classpath configuration for the latest MPS version, which means that the gradle plugin remains compatible with previous MPS version and we don't need to increase the plugin version.

### DIFF
--- a/src/main/groovy/de/itemis/mps/gradle/GenerateLibrariesXml.groovy
+++ b/src/main/groovy/de/itemis/mps/gradle/GenerateLibrariesXml.groovy
@@ -12,7 +12,10 @@ class GenerateLibrariesXml extends DefaultTask {
     @InputFile
     File defaults
 
-    @Optional @InputFile
+    @Optional
+    // currently using @InputFile is not possible due to a failure on non-existing files
+    // TODO: consider using a workaournd with @InputFiles (https://github.com/gradle/gradle/issues/2016)
+    @Input
     File overrides
 
     @OutputFile
@@ -22,11 +25,18 @@ class GenerateLibrariesXml extends DefaultTask {
         description = 'Generates libraries.xml for MPS'
     }
 
+    void setOverrides(Object overrides) {
+        this.overrides = project.file(overrides)
+        if (this.overrides != null && this.overrides.exists()) {
+            inputs.file(overrides)
+        }
+    }
+
     @TaskAction
     def generate() {
         Properties properties = new Properties()
         defaults.withInputStream { properties.load(it) }
-        if (overrides != null && overrides.exists()) {
+        if (overrides.exists()) {
             overrides.withInputStream { properties.load(it) }
         }
         destination.withWriter { writer ->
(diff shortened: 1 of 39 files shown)


src/main/groovy/de/itemis/mps/gradle/GenerateLibrariesXml.groovy
```diff
@@ -12,7 +12,7 @@ class GenerateLibrariesXml extends DefaultTask {
     @InputFile
     File defaults
 
-    @Optional
+    @Optional @InputFile
```
coolya, line 15: I think we can not set this to input file since it will cause an exception if the file doesn't exists. That was the reason why in the past we removed this annotation and introduced the setter method which adds it to the dependencies when file actually exists. 

See the tests that you changed here src/test/kotlin/GenerateProjectLibrariesXml.kt and removed the one that tests with a none existing overrides file.

coolya: Also removing that set method is a breaking API which we can't just do as an MPS upgrade and I would like to avoid.

vlsi: The workaround is to use `@InputFiles`: https://github.com/gradle/gradle/issues/2016

However, I agree it would be better to decouple the change from `MPS upgrade`

wsafonov: It's strange, cause the official documentation states that the task validation should not fail in combination with `@Optional` annotation on `@InputFile` property if the file doesn't exist: https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:task_input_output_validation But you are right - currently the test for non-existing overrides file would fail.

The reason why I actually tried to refactor this is because task validation in validatePlugins task (which we are currently not using in the official build on the build server) failed with a warning due to a property w/o any input or output annotations. Simply using `@Input` annotation doesn't help (still produces a warning, cause `@Input` on Files is deprecated and will be removed in Gradle 7). So `@InputFile` was the obvious choice.

It feels strange from the user point of view that `overrides` unlike other properties requires a setter-style initialization, so I would rather like to refactor this as a breaking change with increased minor number. But given the fact that Gradle currently doesn't really properly support the case with non-existing files, I don't know if it's a good idea. Maybe give a try to the `@InputFiles` workaround if this would behave like a normal property from the user point of view.

So for now I reverted those changes and just set `@Input` annotation as a reminder, extended with some documentation comments.

sergej-koscejev: As a reminder, let's consider dropping this task altogether: #62

coolya (approved):

abstraktor:
Thanks! A client asked us about this just today!

PR #77, opened 2020-09-14
project-loader: reverted change of custom plugins loading

### LINKED ISSUE #76: modelcheck: plugins property processed incorrect
The  modelcheck plugin assumes that all custom plugins configured for the `plugins` property
are placed under a MPS/plugins location which is not required to just load the plugins. 
Instead it should be possible to use the absolute path to determine the location of the plugins.

```
modelcheck {
    plugins = [new Plugin("my.plugin", "my.plugin")]
    ...
}
```
If the custom plugins are not placed under the relative MPS location, 

These changes seem to be wrong: https://github.com/mbeddr/mps-gradle-plugin/commit/16bef1eb2d805b582701df6282739acdc91bd975#r42228923

Loading of custom plugins was supported in the following version of this plugin: 1.2.177.46341c0

arimer:
A list of custom plugins is now added directy to the environment
without the usage of the pre installed plugins path
Fixes #76

### DIFF
--- a/project-loader/src/main/kotlin/Args.kt
+++ b/project-loader/src/main/kotlin/Args.kt
@@ -18,9 +18,10 @@ private fun <T> splitAndCreatePlugin(str: String, creator: (String, String, Bool
         throw RuntimeException("string is not of the right format. Expected <key>::<value>[::<pre installed>]")
     }
     if (split.size == 3){
+        // in case we can't parse true, toBoolean() will always return false
         return creator(split[0], split[1], split[2].toBoolean())
     }
-    return creator(split[0], split[1], true)
+    return creator(split[0], split[1], false)
 }
 
 private fun toMacro(str: String) = splitAndCreate(str, ::Macro)
--- a/project-loader/src/main/kotlin/Args.kt
+++ b/project-loader/src/main/kotlin/Args.kt
@@ -12,8 +12,19 @@ private fun <T> splitAndCreate(str: String, creator: (String, String) -> T): T {
     return creator(split[0], split[1])
 }
 
+private fun <T> splitAndCreatePlugin(str: String, creator: (String, String, Boolean) -> T): T {
+    val split = str.split("::")
+    if (split.size < 2) {
+        throw RuntimeException("string is not of the right format. Expected <key>::<value>[::<pre installed>]")
+    }
+    if (split.size == 3){
+        return creator(split[0], split[1], split[2].toBoolean())
+    }
+    return creator(split[0], split[1], false)
+}
+
 private fun toMacro(str: String) = splitAndCreate(str, ::Macro)
-private fun toPlugin(str: String) = splitAndCreate(str, ::Plugin)
+private fun toPlugin(str: String) = splitAndCreatePlugin(str, ::Plugin)
 
 /**
  * Default set of arguments required to start a "headless" MPS. This class should be used by other users of the
@@ -23,7 +34,7 @@ private fun toPlugin(str: String) = splitAndCreate(str, ::Plugin)
 open class Args(parser: ArgParser) {
 
     val plugins by parser.adding("--plugin",
-            help = "plugin to to load. The format is --plugin=<id>::<path>")
+            help = "plugin to to load. The format is --plugin=<id>::<path>[::<pre installed>]")
     { toPlugin(this) }
 
     val macros by parser.adding("--macro",
--- a/project-loader/src/main/kotlin/Args.kt
+++ b/project-loader/src/main/kotlin/Args.kt
@@ -18,7 +18,7 @@ private fun <T> splitAndCreatePlugin(str: String, creator: (String, String, Bool
         throw RuntimeException("string is not of the right format. Expected <key>::<value>[::<pre installed>]")
     }
     if (split.size == 3){
-        return creator(split[0], split[1], split[2].toBoolean())
+        return creator(split[0], split[1], split[2].toBoolean() ?: true)
     }
     return creator(split[0], split[1], false)
 }
--- a/project-loader/src/main/kotlin/Args.kt
+++ b/project-loader/src/main/kotlin/Args.kt
@@ -20,7 +20,7 @@ private fun <T> splitAndCreatePlugin(str: String, creator: (String, String, Bool
     if (split.size == 3){
         return creator(split[0], split[1], split[2].toBoolean() ?: true)
     }
-    return creator(split[0], split[1], false)
+    return creator(split[0], split[1], true)
 }
 
 private fun toMacro(str: String) = splitAndCreate(str, ::Macro)
--- a/project-loader/src/main/kotlin/Args.kt
+++ b/project-loader/src/main/kotlin/Args.kt
@@ -18,7 +18,7 @@ private fun <T> splitAndCreatePlugin(str: String, creator: (String, String, Bool
         throw RuntimeException("string is not of the right format. Expected <key>::<value>[::<pre installed>]")
     }
     if (split.size == 3){
-        return creator(split[0], split[1], split[2].toBoolean() ?: true)
+        return creator(split[0], split[1], split[2].toBoolean())
     }
     return creator(split[0], split[1], true)
 }
--- a/project-loader/src/main/kotlin/Loader.kt
+++ b/project-loader/src/main/kotlin/Loader.kt
@@ -1,6 +1,5 @@
 package de.itemis.mps.gradle.project.loader
 
-import com.intellij.openapi.application.PathMacros
 import jetbrains.mps.project.Project
 import jetbrains.mps.tool.environment.EnvironmentConfig
 import jetbrains.mps.tool.environment.IdeaEnvironment
@@ -11,7 +10,7 @@ import java.io.File
 data class Plugin(
         val id: String,
         val path: String,
-        val isPreInstalled: Boolean
+        val isCustom: Boolean
 )
 
 data class Macro(
@@ -101,10 +100,10 @@ fun <T> executeWithProject(project: File,
 
     val cfg = basicEnvironmentConfig()
     plugins.forEach {
-        if (it.isPreInstalled) {
-            cfg.addPreInstalledPlugin(it.path, it.id)
-        } else {
+        if (it.isCustom) {
             cfg.addPlugin(it.path, it.id)
+        } else {
+            cfg.addPreInstalledPlugin(it.path, it.id)
         }
     }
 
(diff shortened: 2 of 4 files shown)


coolya:
Does it mean you can only load plugins form the "pluginLocation" with this change or can you still load pre-installed plugins as well?

arimer:
If I understood the current implementation correct, you would need to do both:

```
modelcheck {
  // configure the directoly where "my.plugin" is placed 
    pluginLocation = new File(pluginsDir)
  //create a plugin with correct id  short folder name 
    plugins = [new Plugin("my.plugin", "my.plugin")]
    ...
}
```

coolya:
Thats clear, but can I still load a pre-installed plugin, located into `$mpsHome/plugins`, from the MPS being used with the `plugins` list or will the env only search for them in the `pluginLocation`?

arimer:
>can I still load a pre-installed plugin, located into $mpsHome/plugins, from the MPS being used with the `plugins  list`

No I don't think so, because now the path to the `plugins` (property) list is no longer relative to the `$mpsHome/plugins`

>will the env only search for them in the pluginLocation
Yes

coolya:
> No I don't think so, because now the path to the plugins (property) list is no longer relative to the $mpsHome/plugins

So the `path` property on the plugin now contains what? Before it was the subfolder to search for the plugin descriptor either located in the pre installed plugin location `$mpsHome/plugins` or `$pluginLocation`. So the internal logic where it searches for plugins has changed?

project-loader/src/main/kotlin/Loader.kt
```diff
@@ -100,7 +100,7 @@ fun <T> executeWithProject(project: File,
 
     val cfg = basicEnvironmentConfig()
 
-    plugins.forEach { cfg.addPreInstalledPlugin(it.path, it.id) }
+    plugins.forEach { cfg.addPlugin(it.path, it.id) }
```
arimer, line 103: @coolya Sorry but tbh. I can't follow entirely. The only thing this PR is changing is how the plugins configured in the `plugins` property are loaded. 
Before this commit the project loader was expected them to be placed under `$MPS_HOME/plugins`. 
Jugding from the past, this was not required for the mps-gradle-plugin in version  `1.2.177.46341c0`.

However if the privious implementation was correct, we should then at least update the documentation so that users know where to place their custom plugin. And this location would then always be `$MPS_HOME/plugins`.

coolya: I know that it is changing it, but there was a reason why this change was introduced by Wladi, he was unable to load pre installed plugins like the http support. This was most probably required because of a change in platform where many things moved from relative to absolute paths I don't recall all the details but there was a PR where this is discussed. 

So the question I want to get answered is: 

- is the path property in the plugin always relative to the `pluginLocation`? 
- is the path property in the plugin a absolute path? 

if either of the above is true we need to have a way to load pre installed plugins via the `plugins` list.

arimer: > is the path property in the plugin a absolute path?
If I check the MPS sources for 2019.3 I can see that `env.addPlugin()` is using an absolut path for the path property. see: https://github.com/JetBrains/MPS/blob/2019.3/core/tool/environment/source_gen/jetbrains/mps/tool/environment/EnvironmentConfig.java#L68

coolya: Ok I see, so `pluginLocation` is unused now. Then this PR needs changes to address the fact that plugins are always referenced with their absolute path. 

IMHO we need a way to declare a plugin is "pre installed" in the `plugins` list depending on that the logic need to call `addPreInstalledPlugin` for pre installed ones and `.addPlugin` for not pre installed plugins. In order construct the correct path for plugin in it needs to use `pluginLocation` as a prefix to the `path` defined in the `Plugin` of the `plugins` list otherwise we would need to make all plugin paths absolute which is infeasible because the caller of the plugin might now have control about the location where MPS is extracted.

coolya (changes requested):
See my inline comment

project-loader/src/main/kotlin/Loader.kt
arimer: Agree

arimer:
If I understand your suggestions correct, we could just update this class: 
https://github.com/mbeddr/mps-gradle-plugin/blob/23e933e85f68b0d9091e32959fde62db5ad257f3/project-loader/src/main/kotlin/Loader.kt#L11
with a flag indicating if the pluging is "pre installed" or "custom" right?

project-loader/src/main/kotlin/Loader.kt
```diff
@@ -10,7 +10,8 @@ import java.io.File
 
 data class Plugin(
         val id: String,
-        val path: String
+        val path: String,
+        val isPreInstalled: Boolean?=null
```
coolya, line 14: Why is this a nullable? Wouldn't it make more sense to make it a normal bool and make it false by default? This would allows us to get rid of the null checks below.

project-loader/src/main/kotlin/Args.kt
```diff
@@ -12,8 +12,19 @@ private fun <T> splitAndCreate(str: String, creator: (String, String) -> T): T {
     return creator(split[0], split[1])
 }
 
+private fun <T> splitAndCreatePlugin(str: String, creator: (String, String, Boolean) -> T): T {
+    val split = str.split("::")
+    if (split.size < 2) {
+        throw RuntimeException("string is not of the right format. Expected <key>::<value>[::<pre installed>]")
+    }
+    if (split.size == 3){
+        return creator(split[0], split[1], split[2].toBoolean())
```
coolya, line 21: we could use the "elvis operator" here to check the parse result and otherwise just make preinstalled false? Like this: 
```suggestion
        return creator(split[0], split[1], split[2].toBoolean() ?: false)
```

That was the last argument doesn't have to be nullable.

coolya (commented):
Looks good so far. I think we can go ahead an change the gradle side that calls this code.

project-loader/src/main/kotlin/Loader.kt
arimer: I am not soo much familiar with Kotlin. I though we could make this property optional, so users will not need to update their checkmodels configuration after switching to a newer version.

project-loader/src/main/kotlin/Args.kt
```diff
@@ -12,8 +12,19 @@ private fun <T> splitAndCreate(str: String, creator: (String, String) -> T): T {
...
+    }
+    return creator(split[0], split[1], false)
```
arimer, line 23: With your suggestions we could also get rid of this line 23 and the `if(split.size == 3)` statement I guess.

coolya: This depends how you would like to approach the complete solution. What you changed so far is the internal data structures of the project loader. These data structures are used only by the project loader and aren't what the user puts into their gradle files. 

The project loader is part of a command line application that runs the desired code inside of MPS. The architecture of the mps gradles plugin never runs MPS inside the JVM that is actually executing the gradle build. The mps gradle plugin starts a new JVM/process for all of it MPS related code. The project loader is responsible for setting up the MPS environment and then delegates execution to the action that should be performed on the project, e.g. the model check. The project loader has a command line interface to configure the MPS environment incl. the project to open. 

What you changes so far is the command line interface but not the part that is what user uses in their gradle files. The part that is responsible for the gradle side is in `ModelcheckMpsProjectPlugin`. You can see there that is passes command line arguments to the new JVM its starting which also include the plugins. The class `BasePluginExtensions` contains all the common arguments to configure the MPS environment. In fact it's essentially the very same as in the `Args` class but tailored for use inside of gralde scripts. 

The parsing code with the split etc. that you changed is used parse the command line arguments. Keeping that backward compatible is fine and makes a lot of sense because not every is using gradle as a front end to start the project loader. There are for instance usecases where the project loader / modelchecker are called from maven. 

The next step would be to change the `BasePluginExtensions` to include the "preinstalled" flag and adapt `argsFromBaseExtension` to include it into the command line arguments.

coolya:
On general thing I just noticed, since "pre installed" was the accidentally the default in the past we should also make it the default for the data structures if the argument is missing or cannot be parsed.

project-loader/src/main/kotlin/Args.kt
coolya:
```suggestion
    return creator(split[0], split[1], true)
```

arimer:
Appretiate the explanation very much! I was still unsure about the end user facing part but it is clear now. Thx 👍

coolya:
I think we/I should write this down with a nice diagram to make it easier to understand the code base. I will put it in my todo. If you have more questions feel free to ask.

arimer:
>The next step would be to change the BasePluginExtensions to include the "preinstalled" flag
And there we probably want to make the flag optional, right so that the `modelcheck` config does not need to be adapted I guess.

arimer:
After some discussions with @wsafonov we decided to invert the logic of the new flag and indicate if a plugin is `custom` or not.

wsafonov (changes requested):
My thoughts on it:
- we currently have the `pluginLocation` property which purpose is to specify the path where additional plugins should be loaded from. Usually, IDEA platform uses plugins location in the user specific config directory (like `.MPS2019.3/config/plugins` on Windows) to load additional plugin. But using that property we can override it and let IDEA look for our custom plugin from a custom directory. This actually looks like a good fit for the scenario with a custom plugin. The only issue with it, even if IDEA scans the custom directory and finds our custom plugin there, the loading of plugins is still controlled by the `idea.load.plugins.id` system property which is automatically configured in `jetbrains.mps.tool.environment.IdeaEnvironment` based on the configured plugins in `EnvironmentConfig`. The catch is, to add a plugin to it, the current API requires that the plugin path has to be absolute (we do this currently manually as if all plugins were pre-installed). This renders the `pluginLocation` property essentially useless, as it depends on the list of plugin ids in the `idea.load.plugins.id` property and plugin id from our config would not be added there unless we specify the absolute path for a plugin. So first of all in my opinion we get rid of the `pluginLocation` property as it's effectively not functional anymore and would only confuse the users (or at least mark it as deprecated).
- instead of introducing additional flag for marking plugin location as custom, we could simply interpret non-absolute path as pre-installed plugin (this is anyway the case now) and use the already existing function for building an absolute path for it, and for absolute path just pass it on as it is. Of course, this should be documented in the plugin help as well.

arimer:
closing in favor to #80

PR #82, opened 2020-10-30
update to gradle 6.7

coolya:


sergej-koscejev (approved):

coolya:
I forgot to update some dependency locks in the subprojects thats why the build if failing. Will update them now.

PR #96, opened 2021-04-01
generate, modelcheck: allow using custom MPS

coolya:
Adds support for the custom distributions of MPS vai new property
mpsVersion which is used to resolve the runtime dependency.

Cleanup the code responsible for resolving MPS and removed the logic
to handle an edge case in 2019.2. This causes the plugin to be no longer
compatible with MPS versions lower than 2019.3. The plugin checks the MPS
version and presents an error is its older than 2020.1. The minor version
of the plugin was increased because of that.

Tests were added that contain test project for modelcheck and generate
tasks. One test is ignored because resoving the runtime dependency is still
tricky and not 100% solved. Never the less I wanted to get the feature into
review this week. We can merge it without that test case for now and
revamp the versioning of the runtime part later on.

The readme contains an overview how users can use their custom MPS.

### DIFF
--- a/src/main/kotlin/de/itemis/mps/gradle/Common.kt
+++ b/src/main/kotlin/de/itemis/mps/gradle/Common.kt
@@ -4,9 +4,12 @@ import org.apache.log4j.Logger
 import org.gradle.api.GradleException
 import org.gradle.api.JavaVersion
 import org.gradle.api.artifacts.Configuration
+import org.gradle.initialization.GradleLauncher
 import java.io.File
 
 private val logger = Logger.getLogger("de.itemis.mps.gradle.common")
+
+const val MPS_SUPPORT_MSG = "Version 1.5 doesn't only support MPS 2020.1+, please use versions 1.4 or below with older versions of MPS."
 
 data class Plugin(
         var id: String,
@@ -19,8 +22,9 @@ data class Macro(
 )
 
 open class BasePluginExtensions {
-    lateinit var mpsConfig: Configuration
+    var mpsConfig: Configuration? = null
     var mpsLocation: File? = null
+    var mpsVersion: String? = null
     var plugins: List<Plugin> = emptyList()
     var pluginLocation: File? = null
     var macros: List<Macro> = emptyList()
@@ -48,4 +52,28 @@ fun argsFromBaseExtension(extensions: BasePluginExtensions): MutableList<String>
             extensions.plugins.map { "--plugin=${it.id}::${it.path}" }.asSequence(),
             extensions.macros.map { "--macro=${it.name}::${it.value}" }.asSequence(),
             prj).flatten().toMutableList()
+}
+
+fun BasePluginExtensions.getMPSVersion(): String {
+    /*
+    If the user supplies a MPS config we use this one to resolve MPS and get the version. For other scenarios the user
+    can supply mpsLocation and mpsVersion then we do not resolve anything and the users build script is responsible for
+    resolving a compatible MPS into th mpsLocation before the
+     */
+    if(mpsConfig != null) {
+        return mpsConfig!!
+            .resolvedConfiguration
+            .firstLevelModuleDependencies.find { it.moduleGroup == "com.jetbrains" && it.moduleName == "mps" }
+            ?.moduleVersion ?: throw GradleException("MPS configuration doesn't contain MPS")
+    }
+
+    if(mpsVersion != null) {
+        if(mpsLocation == null) {
+            throw GradleException("Setting an MPS version but no MPS location is not supported!")
+        }
+        return mpsVersion!!
+    }
+
+    throw GradleException("Either mpsConfig or mpsVersion needs to specified!")
+
 }
--- a/src/main/kotlin/de/itemis/mps/gradle/generate/Plugin.kt
+++ b/src/main/kotlin/de/itemis/mps/gradle/generate/Plugin.kt
@@ -1,11 +1,10 @@
 package de.itemis.mps.gradle.generate
 
-import de.itemis.mps.gradle.BasePluginExtensions
-import de.itemis.mps.gradle.argsFromBaseExtension
-import de.itemis.mps.gradle.validateDefaultJvm
+import de.itemis.mps.gradle.*
 import org.gradle.api.GradleException
 import org.gradle.api.Plugin
 import org.gradle.api.Project
+import org.gradle.api.Task
 import org.gradle.api.artifacts.Configuration
 import org.gradle.api.tasks.Copy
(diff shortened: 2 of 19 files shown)


src/main/kotlin/de/itemis/mps/gradle/generate/Plugin.kt
```diff
@@ -28,48 +27,25 @@ open class GenerateMpsProjectPlugin : Plugin<Project> {
...
-                                } else {
-                                    byteArray = outputStream.toByteArray()
-                                }
-                                resultEntries.add(Pair(nextEntry.name, byteArray))
-                            } while (true)
-                            zipTo(File(mpsLocation, this.path), resultEntries.asSequence())
-                        }
-                        exclude()
+                val resolveMps: Task = if(extension.mpsConfig != null) {
+                    tasks.create("resolveMpsForGeneration", Copy::class.java) {
```
vlsi, line 43: `tasks.register`?

src/main/kotlin/de/itemis/mps/gradle/generate/Plugin.kt
```diff
@@ -28,48 +27,25 @@ open class GenerateMpsProjectPlugin : Plugin<Project> {
...
+                        from(extension.mpsConfig!!.resolve().map { zipTree(it) })
+                        into(mpsLocation)
                     }
-                    into(mpsLocation)
+                } else {
+                    tasks.create("resolveMpsForGeneration")
```
vlsi, line 48: `tasks.register`?

src/main/kotlin/de/itemis/mps/gradle/generate/Plugin.kt
```diff
@@ -28,48 +27,25 @@ open class GenerateMpsProjectPlugin : Plugin<Project> {
             val mpsLocation = extension.mpsLocation ?: File(project.buildDir, "mps")
 
             afterEvaluate {
-                val mpsVersion = extension
-                        .mpsConfig
-                        .resolvedConfiguration
-                        .firstLevelModuleDependencies.find { it.moduleGroup == "com.jetbrains" && it.moduleName == "mps" }
-                        ?.moduleVersion ?: throw GradleException("MPS configuration doesn't contain MPS")
+                val mpsVersion = extension.getMPSVersion()
```
vlsi, line 30: What do you think of removing `afterEvaluate` and executing the logic only in case it would be needed in the relevant task?

For instance, `./gradlew help` does not need to compute MPS version.

src/main/kotlin/de/itemis/mps/gradle/Common.kt
```diff
@@ -19,8 +22,9 @@ data class Macro(
 )
 
 open class BasePluginExtensions {
-    lateinit var mpsConfig: Configuration
+    var mpsConfig: Configuration? = null
```
vlsi, line 25: `Property<Configuration>`?

src/main/kotlin/de/itemis/mps/gradle/Common.kt
```diff
@@ -19,8 +22,9 @@ data class Macro(
...
     var mpsLocation: File? = null
+    var mpsVersion: String? = null
```
vlsi, line 27: `Property<String>` ?

src/main/kotlin/de/itemis/mps/gradle/modelcheck/Plugin.kt
```diff
@@ -31,24 +30,24 @@ open class ModelcheckMpsProjectPlugin : Plugin<Project> {
...
-                        ?.moduleVersion ?: throw GradleException("MPS configuration doesn't contain MPS")
+                val mpsVersion = extension.getMPSVersion()
 
                 // this dependency will never resolve against SNAPSHOT version, if there is a previously released version for $mpsVersion
                 // hence if testing SNAPSHOT version locally, replace '+' with '.[pluginVersion]-SNAPSHOT', e.g. '.2-SNAPSHOT'
                 // to make sure that your local SNAPSHOT version will be resolved here
                 val dep = project.dependencies.create("de.itemis.mps:modelcheck:$mpsVersion+")
                 val genConfig = configurations.detachedConfiguration(dep)
 
+                if(mpsVersion.substring(0..3).toInt() < 2020) {
```
vlsi, line 41: `mpsVersion < "2020"` would be easier to read

src/main/kotlin/de/itemis/mps/gradle/modelcheck/Plugin.kt
```diff
@@ -64,36 +63,17 @@ open class ModelcheckMpsProjectPlugin : Plugin<Project> {
...
-                                }
-                                resultEntries.add(Pair(nextEntry.name, byteArray))
-                            } while (true)
-                            zipTo(File(mpsLocation, this.path), resultEntries.asSequence())
-                        }
-                        exclude()
+
+                val resolveMps: Task = if (extension.mpsConfig != null) {
+                    tasks.create("resolveMpsForModelcheck", Copy::class.java) {
+                        from(extension.mpsConfig!!.resolve().map { zipTree(it) })
```
vlsi, line 69: This downloads MPS at the configuration time :(
In other words, it would download MPS even for `./gradlew help`.

I would suggest something like the following

```kotlin
dependsOn(mpsConfig)
from(zipTree(provider { mpsConfig.singleFile }))
```

src/main/kotlin/de/itemis/mps/gradle/modelcheck/Plugin.kt
```diff
@@ -64,36 +63,17 @@ open class ModelcheckMpsProjectPlugin : Plugin<Project> {
...
-                                    byteArray = outputStream.toByteArray()
-                                }
-                                resultEntries.add(Pair(nextEntry.name, byteArray))
-                            } while (true)
-                            zipTo(File(mpsLocation, this.path), resultEntries.asSequence())
-                        }
-                        exclude()
+
+                val resolveMps: Task = if (extension.mpsConfig != null) {
+                    tasks.create("resolveMpsForModelcheck", Copy::class.java) {
```
vlsi, line 68: `Sync` task would work better here. `Copy` won't remove stale files, and `Sync` would remove files as they are removed from the source.

src/main/kotlin/de/itemis/mps/gradle/generate/Plugin.kt
coolya: Nice catch. I'm gonna create a separate PR that removes `afterevalutate` for all other plugins as well. Then we ship this change as a whole in 1.5.

joao-silveira (approved):

PR #105, opened 2021-09-10
build: add github packages publication for artifacts

nkoester:
This PR adds publication into ghp


nkoester:
Not sure why tests fail - as far as i can see it theses are of no relevance to my changes in the build file/package publication?

arimer (approved):
Tbh, I don't entirely understand why the github workflow build is failing. To me it seems like the github actions build is failing for quite a while now, see build of last commit on master: https://github.com/mbeddr/mps-gradle-plugin/runs/3179833086

However the TC build is green and all tests, which are suppose to run are green.

arimer:
I would propose the following: Let's add the ghp publication as it was there in the initial commit and make sure that all tests run green on CI. Afterwards we should create a follow up ticket to fix the breaking tests via github actions. Would this work for you?

nkoester:
This was essentially merged in #106

PR #108, opened 2021-09-30
Remove chmod command on *.py files during MacOS bundle

edenlin18:
Seems like *.py files no longer exist in `$CONTENTS"/bin/`..
Need to get rid of this command or else the build would fail.


sergej-koscejev (changes requested):
I would like the change to be backwards compatible, i.e. guarded by the existence of `*.py`. E.g. similar to the block above that's dealing with jnilib symlinks (`if ls "$CONTENTS/bin"/*.jnilib ...`)

sergej-koscejev:
Declining in favor of #109.

PR #110, opened 2021-10-12
Publish modelcheck and project-loader to GHP

sergej-koscejev:
Centralize publishing repositories configuration in the root project so
that all projects use the same repositories.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -97,26 +97,27 @@ tasks {
     }
 }
 
-publishing {
-    repositories {
-        maven {
-            name = "itemis"
-            url = uri("https://projects.itemis.de/nexus/content/repositories/mbeddr")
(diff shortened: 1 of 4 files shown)


arimer:
Btw. it looks like our github actions tests are rendomly failing. Having a green TC build would be sufficiant from my point of view.

sergej-koscejev:
> Btw. it looks like our github actions tests are rendomly failing.

I believe they are failing because GitHub can't access itemis Nexus. I propose we use GitHub Packages in these tests. One downside of this however is that we would either require GHP also for local testing (more complicated for the developers), or we would need to configure repository differently based on whether we're running on GitHub or locally (complicates the test code and means different test code is run on GitHub versus locally).

What do you think? @arimer @coolya @nkoester ?

PR #118, opened 2022-01-02
jbrdownload: handle new jbr flavours

coolya:
From version 11_0_10+ we published the new falvor scheme form JetBrains
into our nexus. That means the default "jbr" artifact is no longer found
for these versions.
The plugin now handles these versions correctly and optionally allows to
specify the flavour in its configuration.

The plugin now also handles aarch64 cpu architecture for new m1 macs.

### DIFF
--- a/src/main/kotlin/de/itemis/mps/gradle/downloadJBR/Plugin.kt
+++ b/src/main/kotlin/de/itemis/mps/gradle/downloadJBR/Plugin.kt
@@ -9,6 +9,7 @@ import java.io.File
 
 open class DownloadJbrConfiguration {
     lateinit var jbrVersion: String
+    var distributionType : String? = null
     var downloadDir: File? = null
 }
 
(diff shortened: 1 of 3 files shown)


src/main/kotlin/de/itemis/mps/gradle/downloadJBR/Plugin.kt
```diff
@@ -21,15 +22,33 @@ open class DownloadJbrProjectPlugin : Plugin<Project> {
             afterEvaluate {
                 val downloadDir = extension.downloadDir ?: File(buildDir, "jbrDownload")
                 val version = extension.jbrVersion
+                // from version 10 on we the jbr distribution type is replaced with jbr_jcef
```
sergej-koscejev, line 25: Typo here?
```suggestion
                // from version 10 on the jbr distribution type is replaced with jbr_jcef
```

sergej-koscejev (changes requested):
We should bump the minor version of the plugin because we're adding a new feature.

coolya:
> We should bump the minor version of the plugin because we're adding a new feature.

Nice catch totally forgot that. Will do after #122 is merged to avoid a merge conflict.

sergej-koscejev:
> Will do after #122 is merged to avoid a merge conflict.

To be clear: the version I bumped in #122 belongs to the modelcheck backend, not the plugin itself. That PR is also targeting the mps20203 branch whereas this one is for master.

joswarmer:
Did not see this PR, just opened a new issue for this (123), as I currently have a problem with this.

coolya:
> We should bump the minor version of the plugin because we're adding a new feature.

Done.

sergej-koscejev (approved):

PR #119, opened 2022-01-20
update to MPS 2020.3.6

jonaskraemer:
Update for MPS minor version change from 2020.3.5 to 2020.3.6.


arimer:
Changes look good to me. Thanks for the effort 👍🏼

arimer (approved):

jonaskraemer:
There still seems to be a problem with the tests performed via github actions, but on TC and locally the corresponding tests run green, so I assume it's a github issue. Based on old Slack threads I've found this seems to be an expected problem and nothing wrong with this particular PR.

PR #121, opened 2022-01-25
Build: Adapt ghp publication changes to kts

jonaskraemer:
Takes the addition of a publication to ghp from master in order to make the plugins for MPS 2020.3(.6) to available there.


arimer (approved):
Thanks for porting the additional ghp publication back.

PR #122, opened 2022-02-02
Get checkers to run from CheckerRegistry

sergej-koscejev:
Bump plugin version because this may be a breaking change.

This will allow using custom checkers loaded from a customer-specific plugin.


coolya (approved):
looks good to me.

PR #128, opened 2022-03-09
Migrate/migrate mps gradle plugin gradle7.3#116

dbinkele:
Builds are OK:

https://build.mbeddr.com/buildConfiguration/Mbeddr2_Mbeddr_Gradle_MpsGradlePlugin?branch=migrate%2FMigrateMPSGradlePluginGradle7.3%23116&mode=builds


src/main/kotlin/de/itemis/mps/gradle/downloadJBR/Plugin.kt
```diff
@@ -5,12 +5,26 @@ import org.gradle.api.Plugin
...
 import java.io.File
+import javax.inject.Inject
 
-open class DownloadJbrConfiguration {
-    lateinit var jbrVersion: String
-    var distributionType : String? = null
-    var downloadDir: File? = null
+open class DownloadJbrConfiguration @Inject constructor(of: ObjectFactory) {
+    @get:Input
+    val jbrVersion: Property<String>  = of.property(String::class.java)
```
sergej-koscejev, line 19: Same issue: this is an extension, I don't think its properties should be annotated with `@Input` and the like.

src/main/kotlin/de/itemis/mps/gradle/generate/Plugin.kt
```diff
@@ -6,41 +6,47 @@ import org.gradle.api.Plugin
...
 
 open class GenerateMpsProjectPlugin : Plugin<Project> {
-
     override fun apply(project: Project) {
         project.run {
             val extension = extensions.create("generate", GeneratePluginExtensions::class.java)
 
             afterEvaluate {
-                val mpsLocation = extension.mpsLocation ?: File(project.buildDir, "mps")
+                val mpsLocation = extension.mpsLocation.map { it.asFile }.getOrElse(File(project.buildDir, "mps"))
```
sergej-koscejev, line 34: I think this should be solved using a convention for the mpsLocation property.

src/main/kotlin/de/itemis/mps/gradle/generate/Plugin.kt
```diff
@@ -60,15 +66,15 @@ open class GenerateMpsProjectPlugin : Plugin<Project> {
                 *
                 */
                 val fake = tasks.create("fakeBuildNumber", FakeBuildNumberTask::class.java) {
-                    mpsDir = mpsLocation
+                    mpsDir.set(mpsLocation)
```
sergej-koscejev, line 69: if the mpsLocation property gets a convention as I suggested above then we can probably use `extension.mpsLocation` here directly.

src/main/kotlin/de/itemis/mps/gradle/generate/Plugin.kt
```diff
@@ -79,8 +85,8 @@ open class GenerateMpsProjectPlugin : Plugin<Project> {
                     // git4idea: has to be on classpath as bundled plugin to be loaded (since 2019.3)
                     classpath(fileTree(File(mpsLocation, "/plugins")).include("git4idea/**/*.jar"))
                     classpath(genConfig)
-                    debug = extension.debug
-                    main = "de.itemis.mps.gradle.generate.MainKt"
+                    debug = extension.debug.getOrElse(false)
```
sergej-koscejev, line 88: Should probably use convention for `extension.debug` as well?

sergej-koscejev (approved):
I think this will work as it is so I'm approving the PR but I have questions.

src/main/kotlin/de/itemis/mps/gradle/Common.kt
```diff
@@ -1,57 +1,97 @@
...
-    var plugins: List<Plugin> = emptyList()
-    var pluginLocation: File? = null
-    var macros: List<Macro> = emptyList()
-    var projectLocation: File? = null
-    var debug = false
-    var javaExec: File? = null
+open class BasePluginExtensions @Inject constructor(of: ObjectFactory) {
+    @get:Input
+    @get:Optional
+    val mpsConfig: Property<Configuration> = of.property(Configuration::class.java)
```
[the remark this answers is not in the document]
dbinkele: Have look here https://docs.gradle.org/current/userguide/lazy_configuration.html#creating_property_provider

If you are not doing it gradle 7 will complain. The same is true for the optional-annotation.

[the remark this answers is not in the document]
sergej-koscejev: The example there is a task and I agree the annotations are needed for tasks. I don't see where the documentation says that it applies to all properties, can you quote that bit?

src/main/kotlin/de/itemis/mps/gradle/generate/Plugin.kt
dbinkele: The 'getOrElse(File(project.buildDir, "mps"))'-part should be exactly for this. Isn't that the same behavior as in the original code?

sergej-koscejev: I refer to https://docs.gradle.org/current/userguide/lazy_configuration.html#applying_conventions -- this is the standard mechanism to apply default values to lazy properties. But we can leave this for later.

src/main/kotlin/de/itemis/mps/gradle/Common.kt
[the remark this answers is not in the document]
dbinkele: To me this made the impression that these annotations should be used whenever you use a property. Now I only left them if they are inside tasks. Seems to work, but where is that mentioned? Do you know?

src/main/kotlin/de/itemis/mps/gradle/generate/Plugin.kt
dbinkele: Where possible I used the 'convention'. But sometimes I need access to 'project', then it was not possible.

src/main/kotlin/de/itemis/mps/gradle/generate/Plugin.kt
dbinkele: done

src/main/kotlin/de/itemis/mps/gradle/Common.kt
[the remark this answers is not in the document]
sergej-koscejev: I have only seen these annotations documented in the context of Gradle tasks, for example here: https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:up_to_date_checks

sergej-koscejev (commented):
Tests need to be updated. Some "failure expected" tests in ModelCheckTests fail not because of the intended failure but probably due to syntax errors caused by the move to properties.

Documentation needs to be updated to show the new properties-based syntax.

dbinkele:
> 

Done

src/main/kotlin/de/itemis/mps/gradle/generate/Plugin.kt
dbinkele: Moving to properties is what @coolya suggested, which also decreses built-times due to its lazy behavior. Maybe we three should have a short meeting this topic ....

sergej-koscejev: > Maybe we three should have a short meeting this topic

Probably. Also it would be better if we could split the PR, upgrade to Gradle 7.4 in this PR and leave properties for another one.

> I injected 'Project', but I do not know how to get rid of 'mpsDir.set(mpsLocation)' without changing 'FakeBuildNumberTask' quite drastically ....

I think you could then directly use `mpsDir.set(extension.mpsLocation)`.

I have tried this locally and found that adding a convention to mpsLocation means that we can no longer complain if the user sets only mpsVersion but no mpsLocation, so this changes the behavior a bit from what is documented. Not sure if we want that. If not, then your original solution without conventions is fine.

src/main/kotlin/de/itemis/mps/gradle/Common.kt
```diff
@@ -1,57 +1,80 @@
...
-    var pluginLocation: File? = null
-    var macros: List<Macro> = emptyList()
-    var projectLocation: File? = null
-    var debug = false
-    var javaExec: File? = null
+open class BasePluginExtensions @Inject constructor(of: ObjectFactory, project: Project) {
+
+    val mpsConfig: Property<Configuration> = of.property(Configuration::class.java)
+
+    val mpsLocation: RegularFileProperty = of.fileProperty().convention { File(project.buildDir, "mps") }
```
sergej-koscejev, line 42: Can inject `layout: ProjectLayout` in the constructor instead of `project` and change this property into a `DirectoryProperty` like this:
```
    val mpsLocation: DirectoryProperty = of.directoryProperty().convention(layout.buildDirectory.dir("mps"))
```

This will make it easier to work with this property later on, e.g. when computing the `plugins` or `lib` directory.

PR #132, opened 2022-08-31
 build: added publication to new nexus itemis cloud

arimer:
**Reasons for changes**
- Some of our downstream projects depent on mps20203 artifacts which are not yet published to the itemis cloud.
- In particular  the 2020.3 version of [iets3.os](https://github.com/IETS3/iets3.opensource/tree/maintenance/mps20203) is depending on `de.itemis.mps:mps-gradle-plugin:1.2.175.cc60dc8`
-  that artifact needs to be available under https://artifacts.itemis.cloud/repository/maven-mps/

This PR is adding the missing publication part to the new itemis cloud.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -53,6 +53,7 @@ repositories {
     mavenCentral()
     maven {
         url = URI("https://projects.itemis.de/nexus/content/repositories/mbeddr")
+        url = URI("https://artifacts.itemis.cloud/repository/maven-mps/")
     }
 }
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -104,11 +104,7 @@ allprojects {
         repositories {
             maven {
                 name = "itemisCloud"
-                url = if(project.hasProperty("useSnapshot")) {
-                        uri("https://artifacts.itemis.cloud/repository/maven-mps-snapshots/")
-                      } else { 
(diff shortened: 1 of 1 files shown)


alexanderpann (approved):

arimer:
@alexanderpann in addition I decided to remove the publicaiton to the old nexus, because in any case it will be shut down soon.

sergej-koscejev:
I don't think we need to support snapshots, we also didn't support them before.

I would appreciate if you also cherry-pick this to `v1.x` branch. Or I can do it myself when I work on the plugin again.

sergej-koscejev:
I believe 1.9.x should be backwards compatible with 1.2.x but the best way to find out is to try...

sergej-koscejev (approved):

arimer:
>I  believe 1.9.x should be backwards compatible with 1.2.x but the best way to find out is to try...

After this merge I should have 1.6.x in the repo. Which hopefully should be compatible with 1.2.

>I don't think we need to support snapshots, we also didn't support them before.

Did not know that. I just added the very same changes which are applied on master. How about we remove it on master and merge it back/cherry-pick it to the maintenence branches?

>I would appreciate if you also cherry-pick this to v1.x branch.

Will do after the merge.

arimer:
Nevermind, I removed the snapshot publication in this PR.

PR #134, opened 2022-09-12
Remove old project.itemis nexus dependency

### LINKED ISSUE #103: Build failed on GitHub but runs locally
This build failed on GitHub: https://github.com/mbeddr/mps-gradle-plugin/actions/runs/1074172041

The same code passed the build on TeamCity as part of the #102 PR checks.

Locally `./gradlew test` also passes.

arimer:
This PR removes the dependency to our old NEXUS and switches to the new one: _artifacts.itemis.cloud_
Under this scope some minor version updates were done to match existing artifact versions on the new nexus.

- JDK/JBR version updated to 11_0_10-b1341.41 in tests
- ~Updated MPS version to 2020.3.6 in build and tests~
- Removed snapshot publication and publication to old nexus

Those changes should also fix #103

### DIFF
--- a/README.md
+++ b/README.md
@@ -426,7 +426,7 @@ being set and no value set for `mpsConfig`. If you set `mpsVersion` but also set
 will take precedence over `mpsVersion` and the plugin will resolve that configuration into `mpsLocation`. 
 
 `mpsVersion` needs to be set to the exact MPS version your custom distribution is based on e.g. if you build a
-RCP with MPS 2020.3.6 you need to set this property to `2020.3.6`. `mpsLocation` needs to point to the location
+RCP with MPS 2020.3.4 you need to set this property to `2020.3.4`. `mpsLocation` needs to point to the location
 where you extracted your custom MPS distribution into e.g. `$buildDir/myAwesomeMPS` if you extracted into that location. 
 
 Each of the plugins creates a `resolveMpsFor<name>` task in the build. When `mpsVersion` and `mpsLocation` are set
@@ -443,7 +443,7 @@ task downloadAndExtractCustomMPS() {
 
 modelcheck {
     mpsLocation.set(myCustomLocation)
-    mpsVersion.set("2020.3.6")
+    mpsVersion.set("2020.3.4")
     projectLocation.set(file("$rootDir/mps-prj"))
     modules.set(["my.solution.with.errors"])
     junitFile.set(file("$buildDir/TEST-modelcheck-results.xml"))
--- a/README.md
+++ b/README.md
@@ -386,7 +386,7 @@ repositories {
 }
 
 downloadJbr {
-    jbrVersion.set("11_0_10-b1341.41")
+    jbrVersion.set("11_0_10-b1145.96")
 }
 ```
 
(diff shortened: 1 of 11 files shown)


arimer:
After looking at all the different feature branches `v1.x`, `mps/20203`, `mps/20211` I am no longer sure to which target branch my PR should go. 
I'd appriciate any explanation.

wsafonov (changes requested):
In generall looks good, just minor stuff: 
- we update here to the 2020.3.6 MPS version, which was packaged with the JBR version `11_0_10-b1145.96`. We should probably stick to this version to stay in sync with the JBR version used in MPS
- we can now remove the old nexus repo from publishing repos in this PR as well, as the old nexus anyway gets freezed in couple of days

wsafonov:
> After looking at all the different feature branches `v1.x`, `mps/20203`, `mps/20211` I am no longer sure to which target branch my PR should go. I'd appriciate any explanation.

I'm also confused by the current branch scheme. The old branching scheme sticked to MPS versions, but then there was an effort by Sergej to refactor this to use semantic versions in https://github.com/mbeddr/mps-build-backends. But the branch `v1.x` introduces schemantic versioning in this repo. So now it got kind of messy here with mixing up different schemes. Probably @sergej-koscejev is the best to provide more information on this.

This also now concerns my comment above on JBR version for MPS 2020.3.6 update, as there is now a separate branch `mps20203` for MPS 2020.3.x which should include that change and not `master`. But I'm not sure about the role of the `master` branch in the new scheme, so we better clarify all this before merging this PR.

sergej-koscejev:
The only active branches are master and v1.x. I hope that the others are not used but I don't really know.

arimer:
@wsafonov thx for the review. I missed the relationship between the JBR and the MPS version entirely, nice catch 👍🏼 
In case only `master` and ` v1.x` is active, should we then merge back `mps/20211` into master? Because master is then outdated 🤷🏼

alexanderpann:
> We should probably stick to this version to stay in sync with the JBR version used in MPS

@wsafonov I am just curious: did we ever have a bug related to using a mismatched JBR version? I can't imagine that a higher JBR version introduces any breaking changes that would affect MPS. The only big change in the past that I can think of is the drop of support for JavaFX. Is there anything else that I am missing?

arimer:
Updated: 
- we will remove the MPS related changes in this PR
- we will update our downloadJDK tests to match the availale JDK version + distribution types

arimer:
@wsafonov please have a second look. (I will squash the commits after approval)

wsafonov:
> > We should probably stick to this version to stay in sync with the JBR version used in MPS
> 
> @wsafonov I am just curious: did we ever have a bug related to using a mismatched JBR version? I can't imagine that a higher JBR version introduces any breaking changes that would affect MPS. The only big change in the past that I can think of is the drop of support for JavaFX. Is there anything else that I am missing?

I'm not aware of serious issues with such incompatibility. It's just to keep the configuration in sync with the corresponding MPS version used on the target branch and to keep everything clean. We would need anyway to clarify the branches scheme and release process for different MPS versions later on.

wsafonov (approved):

PR #135, opened 2022-09-14
Fix usage of the distributionType property in downloadJbr plugin

wsafonov:
`distributionType`  property of the downloadJbr plugin was not properly evaluated in case it was provided by the user. This PR fixes that issue.

### DIFF
--- a/src/test/kotlin/test/de/itemis/mps/gradle/JBRDownloadTest.kt
+++ b/src/test/kotlin/test/de/itemis/mps/gradle/JBRDownloadTest.kt
@@ -36,7 +36,7 @@ class JBRDownloadTest {
             import java.net.URI
             buildscript {
                 dependencies {
-                    "classpath"(files(${cp.map { """"${it.invariantSeparatorsPath}"""" }.joinToString() }))
+                    "classpath"(files(${cp.joinToString { """"${it.invariantSeparatorsPath}"""" }}))
                 }
             }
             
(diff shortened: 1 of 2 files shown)


src/test/kotlin/test/de/itemis/mps/gradle/JBRDownloadTest.kt
```diff
@@ -140,7 +140,11 @@ class JBRDownloadTest {
                 .withPluginClasspath(cp)
                 .build()
         Assert.assertEquals(TaskOutcome.SUCCESS, result.task(":downloadJbr")?.outcome)
-        Assert.assertTrue(File(testProjectDir.root, "build/jbrDownload").exists())
+        val jbrDownloadDir = File(testProjectDir.root, "build/jbrDownload")
+        Assert.assertTrue(jbrDownloadDir.exists())
+        val jbrReleaseFile = File(jbrDownloadDir, "jbr/release")
```
arimer, line 145: What is that release directory? I never saw it in any of the donwloaded JBR distros.

arimer: Just cross-checked, it seems like it is only available on the linux version of the JBR

arimer (approved):
Just some smaller remarks in th comments, but it should not hinder us from merging. 

Thx for the fix 💪🏼

src/test/kotlin/test/de/itemis/mps/gradle/JBRDownloadTest.kt
wsafonov: That file is packaged with all platform distros, I was running the tests locally on Windows, which worked. But probably you checked it on OSX - there the JBR directories layout is simply different and therefore the file would not be found in the test. If you wish, you could fix it. But I think, we are good enough with it for now.

PR #137, opened 2023-01-16
Add support for running migrations

alexanderpann:
The new `run-migrations` plugin works similar to the modelcheck plugin:

```
apply plugin: 'run-migrations'

configurations {
    mps
}

dependencies {
    mps "com.jetbrains:mps:$mpsVersion"
}

runMigrations {
    projectLocation.set(new File("./mps-prj"))
    mpsConfig.set(configurations.mps)
}
```

The plugin should work with any MPS version as all jars are included as dependencies. It uses the `migrate` Ant task directly. There is a `makeDistribModules` flag that I haven't implemented because I am not sure if we need it.

One of the classes that this task uses is [MPSLoadTask](https://github.com/JetBrains/MPS/blob/master/core/tool/ant/source_gen/jetbrains/mps/build/ant/MpsLoadTask.java) which might support some more arguments as well as the [MigrateTaskProperties](https://github.com/JetBrains/MPS/blob/master/core/tool/common/source_gen/jetbrains/mps/tool/common/MigrateTaskProperties.java) in newer MPS versions which declares some attributes. Please let me know if we require any of those properties and list their Ant attribute names.
The only examples for invoking this Ant task can be found in [mpsMigrationTest.xml](https://github.com/JetBrains/MPS/blob/master/build/tests/mpsMigrationTest.xml) and [mpsSmartRefAttrMigration.xml](https://github.com/JetBrains/MPS/blob/master/build/tests/mpsSmartRefAttrMigration.xml).

One additional question: is there a trick to save part of a [Groovy builder](https://docs.gradle.org/current/userguide/kotlin_dsl.html#the_kotlin_dsl_groovy_builder) into a variable in a Gradle script? I want to write code like this:

```
val ref = "test"("123")

ant.withGroovyBuilder {   
	"hello"("word")
    ref
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,7 +245,7 @@ Parameters:
 * `mpsConfig` - the configuration used to resolve MPS. Currently only vanilla MPS is supported and no custom RCPs.
   Custom plugins are supported via the `pluginLocation` parameter.
 * `mpsLocation` - optional location where to place the MPS files.
-* `mpsVersion` - optional if you use a [custom distribution](#Custom MPS Distribution) of MPS
+* `mpsVersion` - optional if you use a [custom distribution](#custom-mps-distribution) of MPS
 * `javaExec` - optional `java` executable to use.
 * `pluginLocation` - location where to load the plugins from. Structure needs to be a flat folder structure similar to the
   `plugins` directory inside of the MPS installation.
@@ -299,7 +299,7 @@ Parameters:
 * `mpsConfig` - the configuration used to resolve MPS. Currently only vanilla MPS is supported and no custom RCPs.
   Custom plugins are supported via the `pluginLocation` parameter.
 * `mpsLocation` - optional location where to place the MPS files.
-* `mpsVersion` - optional if you use a [custom distribution](#Custom MPS Distribution) of MPS
+* `mpsVersion` - optional if you use a [custom distribution](#custom-mps-distribution) of MPS
 * `javaExec` - optional `java` executable to use.
 * `pluginLocation` - location where to load the plugins from. Structure needs to be a flat folder structure similar to the
   `plugins` directory inside of the MPS installation.
@@ -330,7 +330,40 @@ Parameters:
 * `maxHeap` - maximum heap size setting for the JVM that executes the modelchecker. This is useful to limit the heap usage
   in scenarios like containerized build agents where the OS reported memory limit is not the maximum
   to be consumed by the container. The value is a string understood by the JVM command line argument `-Xmx` e.g. `3G` or `512M
-  
+
+
+## Run migrations
+
+Run all pending migrations in the project.
+
+### Usage
+
+A minimal build script to check all models in a MPS project with no external plugins would look like this:
+
+```
+apply plugin: 'run-migrations"'
+
+configurations {
+    mps
+}
+
+dependencies {
+    mps "com.jetbrains:mps:$mpsVersion"
+}
+
+runMigrations {
+    projectLocation.set(new File("./mps-prj"))
+    mpsConfig.set(configurations.mps)
+}
+```
+
+Parameters:
+* `mpsConfig` - the configuration used to resolve MPS.
+* `mpsLocation` - optional location where to place the MPS files.
+* `mpsVersion` - optional if you use a [custom distribution](#custom-mps-distribution) of MPS.
+* `projectLocation` - optional location of the project that should be migrated.
+* `force` - ignores the marker files for projects which allow pending migrations, migrate them anyway (supported in 2021.3.0 and higher)
+
 ### Additional Plugins 
 
 By default only the minimum required set of plugins are loaded. This includes base language and some utilities like the
--- a/README.md
+++ b/README.md
@@ -358,11 +358,13 @@ runMigrations {
 ```
 
 Parameters:
-* `mpsConfig` - the configuration used to resolve MPS.
-* `mpsLocation` - optional location where to place the MPS files.
-* `mpsVersion` - optional if you use a [custom distribution](#custom-mps-distribution) of MPS.
-* `projectLocation` - optional location of the project that should be migrated.
+* `mpsConfig` - configuration used to resolve MPS.
+* `mpsLocation` - location where to place the MPS files.
+* `mpsVersion` - if you use a [custom distribution](#custom-mps-distribution) of MPS.
+* `projectLocation` - location of the project that should be migrated.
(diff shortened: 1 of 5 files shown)


sergej-koscejev:
We need this functionality in the v1.x branch. I'll review it tomorrow.

src/main/kotlin/de/itemis/mps/gradle/runmigrations/Plugin.kt
```diff
@@ -0,0 +1,83 @@
...
+    val force: Property<Boolean> = of.property(Boolean::class.java).convention(false)
+}
+@Suppress("unused")
+open class RunMigrationsMpsProjectPlugin : Plugin<Project> {
+    override fun apply(project: Project) {
+        project.run {
+            val extension = extensions.create("runMigrations", MigrationExecutorPluginExtensions::class.java)
+            afterEvaluate {
+                val mpsLocation = extension.mpsLocation.convention{ File(project.buildDir, "mps") }.map { it.asFile }.get()
+                val projectLocation = extension.projectLocation.getOrElse { throw GradleException("No project path set") }
```
sergej-koscejev, line 26: The docs in README.md say projectLocation is optional yet here an error is thrown if it's not set?

src/main/kotlin/de/itemis/mps/gradle/runmigrations/Plugin.kt
```diff
@@ -0,0 +1,83 @@
...
+                val resolveMps: Task = if (extension.mpsConfig.isPresent) {
+                    tasks.create("resolveMpsForMigrations", Copy::class.java) {
+                        from({extension.mpsConfig.get().resolve().map { zipTree(it) }})
+                        into(mpsLocation)
+                    }
+                } else {
+                    tasks.create("resolveMpsForMigrations")
+                }
+                
+                tasks.register("runMigrations") {
```
sergej-koscejev, line 47: I found it's useful to register tasks right when the plugin is applied, before `project.afterEvaluate`, and then configure them in `project.afterEvaluate`.

However, in general, it's best to do as little work as possible in `project.afterEvaluate` because it's called late in the lifecycle. If the user wants to adjust or override something that our plugin does in `afterEvaluate`, they may have to resort to adding their own `project.afterEvaluate` callback and it gets ugly quickly.

src/main/kotlin/de/itemis/mps/gradle/runmigrations/Plugin.kt
```diff
@@ -0,0 +1,83 @@
...
+                                "fileset"("dir" to  "$mpsLocation/lib", "includes" to "**/*.jar")
+                            }
+                            "taskdef"("resource" to "jetbrains/mps/build/ant/antlib.xml", "classpathref" to "path.mps.ant.path")
+                        }
+                        if(forceMigration.getOrElse(false)) {
+                            ant.withGroovyBuilder {
+                                "migrate"("project" to projectLocation, "mpsHome" to mpsLocation, "force" to true) {
+                                    "macro"("name" to "mps_home", "path" to mpsLocation)
+                                    "jvmargs"() {
+                                        "arg"("value" to "-Didea.log.config.file=log.xml")
```
sergej-koscejev, line 62: I think we should make heap size configurable, i.e. let the user specify a value and we convert it to a "-Xmx" argument.

src/test/kotlin/test/de/itemis/mps/gradle/RunMigrationsTest.kt
```diff
@@ -0,0 +1,185 @@
...
+                force.set(true)
+            }
+        """.trimIndent()
+        )
+
+        GradleRunner.create()
+            .withProjectDir(testProjectDir.root)
+            .withArguments("runMigrations")
+            .withPluginClasspath(cp)
+            .buildAndFail()
```
sergej-koscejev, line 108: The failure tests should check that the build failed in the expected place (some such tests passed because the build failed with an unrelated error).

README.md
```diff
@@ -330,7 +330,40 @@ Parameters:
...
+    projectLocation.set(new File("./mps-prj"))
+    mpsConfig.set(configurations.mps)
+}
+```
+
+Parameters:
+* `mpsConfig` - the configuration used to resolve MPS.
+* `mpsLocation` - optional location where to place the MPS files.
+* `mpsVersion` - optional if you use a [custom distribution](#custom-mps-distribution) of MPS.
+* `projectLocation` - optional location of the project that should be migrated.
```
sergej-koscejev, line 364: I think 'optional' here contradicts the actual code.

sergej-koscejev (approved):

src/main/kotlin/de/itemis/mps/gradle/runmigrations/Plugin.kt
```diff
@@ -0,0 +1,83 @@
...
+            val extension = extensions.create("runMigrations", MigrationExecutorPluginExtensions::class.java)
+            afterEvaluate {
+                val mpsLocation = extension.mpsLocation.convention{ File(project.buildDir, "mps") }.map { it.asFile }.get()
+                val projectLocation = extension.projectLocation.getOrElse { throw GradleException("No project path set") }
+                if(!file(projectLocation).exists()) {
+                    throw GradleException("The path to the project doesn't exist:$projectLocation")
+                }
+                val forceMigration = extension.force
+                
+                val mpsVersion = extension.getMPSVersion()
```
sergej-koscejev, line 32: How will this part work with a custom RCP? Will I have to know which version of my RCP corresponds to which version of MPS? I'd rather not to. What happens if we don't do the check and a user uses 'force' on an early version of MPS?

src/main/kotlin/de/itemis/mps/gradle/runmigrations/Plugin.kt
alexanderpann: I've fixed the readme.

README.md
alexanderpann: I've fixed the readme.

src/main/kotlin/de/itemis/mps/gradle/runmigrations/Plugin.kt
alexanderpann: Explanation for others: as long as `mpsLocation` and `mpsVersion` are set, it will work. `mpsVersion` can be set to a random value like 999.999. The only requirement is that [ semantic versioning](https://semver.org/) is used for the version. The MPS version checks should pass this way.

PR #139, opened 2023-01-30
Fix jbr download folder name for JBR 17

alexanderpann:
Starting from JBR 17, the folder is named jbr-[distributionType]-[version] and not just jbr.


alexanderpann:
BTW, I've figure out how to test the plugin locally with a composite build: [snippet](https://gist.github.com/alexanderpann/4c2ea06bc359db69dad50ba01b381fb9 )
It's a lot easier than I thought.

PR #146, opened 2023-09-05
Pass --parallel parameter to checker runner

cdelabou:
Follows up with [mps-build-backends#11](https://github.com/mbeddr/mps-build-backends/pull/11), pass the boolean flag down to the model checker runner


sergej-koscejev:
Implemented as part of #149

PR #147, opened 2023-10-06
runMigrations: add haltOnDependencyError flag

kolmar:


HeikoBecker (changes requested):
The code itself looks fine to me.

I have one request for rewording on the README:

To make sure that there is no confusion for future users of this plugin, can you reword the descriptions of `haltOnPrecheckFailure` and `haltOnDependencyError` to explain what happens when the flags are true? Currently it reads to me as if the flag being set (to true) has the opposite effect than what is actually described in the README.

HeikoBecker (approved):
Thanks for the fixes. README looks good now.

PR #151, opened 2023-12-05
Make downloadJbr accept the distributionType parameter

HeikoBecker:
Previously the parameter was ignored, always downloading jbr or jbr_jcef depending on the version string.


sergej-koscejev:
Please update the version and the change log.

HeikoBecker:
> Please update the version and the change log.

Done.

PR #152, opened 2023-12-06
downloadJbr: add support for Jbr versions with dashes

HeikoBecker:
Previously only jbr versions with an underscore were supported. Now `jbr-17...` works as well.

### DIFF
--- a/src/main/kotlin/de/itemis/mps/gradle/downloadJBR/Plugin.kt
+++ b/src/main/kotlin/de/itemis/mps/gradle/downloadJBR/Plugin.kt
@@ -46,7 +46,7 @@ open class DownloadJbrProjectPlugin : Plugin<Project> {
                             workingDir = downloadDir
                         }
 
-                        if (downloadDir.listFiles { _, name -> name.startsWith("jbr_") }!!.any()) {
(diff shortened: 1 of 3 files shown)


src/main/kotlin/de/itemis/mps/gradle/downloadJBR/Plugin.kt
```diff
@@ -46,7 +46,7 @@ open class DownloadJbrProjectPlugin : Plugin<Project> {
                             workingDir = downloadDir
                         }
 
-                        if (downloadDir.listFiles { _, name -> name.startsWith("jbr_") }!!.any()) {
+                        if (downloadDir.listFiles { _, name -> name.startsWith("jbr_") || name.startsWith("jbr-") }!!.any()) {
                             exec {
                                 commandLine("sh", "-c", "mv jbr_* jbr")
```
sergej-koscejev:
```suggestion
                                commandLine("sh", "-c", "mv jbr* jbr")
```

sergej-koscejev (changes requested):
Need to fix the `mv` command as well.

sergej-koscejev (approved):

PR #154, opened 2024-04-10
Align gradle kotlin version

wjglerum:
Removes obsolete and empty `compileOnly` configuration from the lockfile and aligns the used Kotlin version with the embedded Kotlin version from Gradle.

This is a first attempt to update this plugin, to get rid of vulnerability warnings. And in the long run to make sure we can use the configuration cache in a newer version of Gradle in the future.

https://docs.gradle.org/current/userguide/compatibility.html#kotlin
https://kotlinlang.org/docs/releases.html#release-details


sergej-koscejev:
I'm trying to be careful with Gradle and Kotlin version upgrades because the plugin may easily become incompatible with older Gradle versions (including older than 7.6.x) and not every project upgrades their Gradle regularly. But I'm going to approve this and see what happens.

PR #159, opened 2024-06-22
[breaking] MPS task interface

sergej-koscejev:
### Changed

- All MPS tasks now implement `MpsTask` interface which defines common properties. As a result, some tasks were
  affected (all marked as `@Incubating`):
  - `MpsCheck`: `pluginRoots` type was changed from `SetProperty<Directory>` to `ConfigurableFileCollection` because the latter is easier to work with from Gradle. **This is a breaking change.**
  - `MpsExecute`: `folderMacros` are now supported and will be passed to MPS.

### Deprecated

- `MpsCheck`, `MpsGenerate`, `MpsExecute`: `varMacros` (or `macros` in case of `MpsExecute`) are now deprecated and will be removed in a later release. Only folder macros (pointing to an existing directory) are supported by MPS, any non-folder macros would not have worked properly anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,22 @@
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 1.26.0
+
+### Changed
+
+- All MPS tasks now implement `MpsTask` interface which defines common properties. As a result, some tasks were
+  affected:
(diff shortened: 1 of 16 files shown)


sergej-koscejev:
I created a PR instead of pushing the change to v1.x right away because it's potentially breaking and I would like to gather some opinions on it in case I'm unknowingly making a big mistake.

sergej-koscejev:
Having a common interface makes it easier to configure the common properties for all tasks that launch MPS (migrations, generation, model checking, etc.) using `tasks.withType(MpsTask).configureEach { ... }`.

sergej-koscejev:
I have a different idea now, I think we should introduce `MpsForkOptions` interface for these tasks, similar to `JavaForkOptions` (and probably extending it).

PR #160, opened 2024-09-10
Not only the 'master' prefix but also the 'main' prefix is now stripped when deriving the version identifier

alexanderpann:


sergej-koscejev:
The version in git-based-versioning/build.gradle.kts has to be updated.

PR #161, opened 2024-09-29
Cleanup + move some common functionality from platform build scripts into this plugin

alexanderpann:
Many lines of code always repeat in our build scripts. I exposed a few new methods through extensions and also cleaned up a few things that I found. Please let me know what you think about the changes, as I am not a Gradle expert. I can remove unwanted commits.

For the example build script, the length changes from 294 lines to 187:
https://gist.github.com/alexanderpann/67cc16888f425560990edc2b3f408654


sergej-koscejev:
Hi @alexanderpann, please split this PR into smaller, focused PRs to simplify review.

sergej-koscejev:
In general, conventions and utilities that are specific to the way that itemis libraries are being built should go into a separate Gradle plugin. It may be part of this repository but should probably have its own subproject because it may end up having different dependencies.

sergej-koscejev (changes requested):
Requesting changes as described above

alexanderpann:
If someone else wants to take over this PR, feel free to do so. At the moment, I don't plan to merge this PR.

nkoester:
This PR will not be merged, however we will cherry-pick changes from here in #169

PR #166, opened 2025-05-07
Avoid calling `Task.project`

wjglerum:
See the docs:
- https://docs.gradle.org/8.12.1/userguide/upgrading_version_7.html#task_project
- https://docs.gradle.org/8.12.1/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution

Which suggest to use `Task.logger` directly instead to be compatible with the configuration cache.


wjglerum:
Just tried this with a local build and the warning is gone from the logs

sergej-koscejev (approved):

PR #168, opened 2025-05-21
Remove another usage of `Task.getProject()`

wjglerum:
The docs tell us to only use `Task.logging`, Im not sure what the actual intent of this function was, so I simply removed this bit.

- https://docs.gradle.org/current/userguide/upgrading_version_8.html#task_project
- https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution

### DIFF
--- a/src/main/kotlin/de/itemis/mps/gradle/tasks/backend_arguments.kt
+++ b/src/main/kotlin/de/itemis/mps/gradle/tasks/backend_arguments.kt
@@ -2,24 +2,11 @@ package de.itemis.mps.gradle.tasks
 
 import de.itemis.mps.gradle.ErrorMessages
 import org.gradle.api.GradleException
-import org.gradle.api.Task
 import org.gradle.api.file.Directory
 import org.gradle.api.file.FileCollection
 import org.gradle.api.file.FileSystemLocation
-import org.gradle.api.logging.LogLevel
 import org.gradle.api.provider.Provider
 import java.io.File
-
-internal fun Task.addLogLevel(args: MutableCollection<String>, logLevel: LogLevel) = addIfInfoLogLevel(args, logLevel,"--log-level=info")
-
-/**
- * A weird overload, usable for building a map of Ant task attributes, as well as a list of command line backend
- * attributes.
- */
-internal fun <T> Task.addIfInfoLogLevel(result: MutableCollection<T>, logLevel: LogLevel, value: T) {
-    val effectiveLogLevel = logging.level ?: logLevel
-    if (effectiveLogLevel <= LogLevel.INFO) result.add(value)
-}
 
 internal fun checkProjectLocation(projectLocation: Provider<out FileSystemLocation>) =
     checkProjectLocation(projectLocation.get().asFile)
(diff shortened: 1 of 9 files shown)


sergej-koscejev:
The intent was to enable MPS info logging if Gradle is started with `--info`. I believe this PR would break this. Perhaps we should introduce a logLevel property and use the start parameter value as a convention. (Or could we reuse the Gradle task log level for that?)

wjglerum:
> The intent was to enable MPS info logging if Gradle is started with `--info`. I believe this PR would break this. Perhaps we should introduce a logLevel property and use the start parameter value as a convention. (Or could we reuse the Gradle task log level for that?)

Yeah later realised that too, we somehow have to get rid of this `Task#getProject` call

sergej-koscejev:
We can use the project for convention values, I believe, just not at execution time.

wjglerum:
> We can use the project for convention values, I believe, just not at execution time.

Just updated the PR with a convention value for each task, that seems to work indeed

src/main/kotlin/de/itemis/mps/gradle/tasks/backend_arguments.kt
```diff
@@ -10,14 +10,14 @@ import org.gradle.api.logging.LogLevel
...
-internal fun Task.addLogLevel(args: MutableCollection<String>) = addIfInfoLogLevel(args, "--log-level=info")
+internal fun Task.addLogLevel(args: MutableCollection<String>, logLevel: LogLevel) = addIfInfoLogLevel(args, logLevel,"--log-level=info")
 
 /**
  * A weird overload, usable for building a map of Ant task attributes, as well as a list of command line backend
  * attributes.
  */
-internal fun <T> Task.addIfInfoLogLevel(result: MutableCollection<T>, value: T) {
-    val effectiveLogLevel = logging.level ?: project.gradle.startParameter.logLevel
+internal fun <T> Task.addIfInfoLogLevel(result: MutableCollection<T>, logLevel: LogLevel, value: T) {
```
sergej-koscejev: The two `add*` methods should no longer need `Task` as their receiver. This one could probably even be inlined because it doesn't bring much value anymore, apart from knowing that 'INFO' is somehow a significant log level, but I'm on the fence here.

sergej-koscejev (approved):

src/main/kotlin/de/itemis/mps/gradle/tasks/backend_arguments.kt
wjglerum: We do need the `Task` as a receiver to be able to resolve `logging.level` from the task afaik here

sergej-koscejev: Ah right. But I thought we wanted to get rid of that anyway? I have never actually seen task's logging.level be different from null.

wjglerum: True, I'll simplify it some more 👍

sergej-koscejev:
Looks good (still), please clean up the commit history and rebase on top of #171, I guess? Maybe both changes could be combined in one bugfix release.

wjglerum:
> Looks good (still), please clean up the commit history and rebase on top of #171, I guess? Maybe both changes could be combined in one bugfix release.

I'll combine both into one PR and cleanup the history 👍

wjglerum:
Combined with https://github.com/mbeddr/mps-gradle-plugin/pull/171

Issue #169, opened 2025-05-23
v3 preparations


nkoester:
This issue serves as a collection of tasks to do before a v3 can be release. There has been a previous attempt over in #100. These are further extensions to be considered once this list of tasks is done.

Identified tasks for v3 preparations (in order of prio):

- [x] Use Gradle Version Catalog -> #170 
- [ ] Move to use gradle lazy properties
- [ ] Deprecate/delete modelcheck and generate plugins in favor of MpsCheck and MpsGenerate tasks
- [ ] Update dependency version
     - [ ] Gradle -> #171 
     - [ ] Kotlin / other dependencies
- [ ] Refine or delete the Github Actions -> #172 
- [ ] Analyze and cherry-pick from efforts in this PR #161 
- [ ] Analyze and cherry-pick from `master` (aka. v2)
- [ ] Analyze and implement efforts from #100 
  - [ ] Deprecate the JBR download plugin and related tasks in favor of https://plugins.gradle.org/plugin/com.specificlanguages.jbr-toolchain or some JetBrains-provided solution such as mentioned in https://github.com/mbeddr/mps-gradle-plugin/issues/100.
- [ ] Make sure all tasks use FileCollection for collections of files (even directories, etc., such as plugin roots) rather than SetProperty<Directory> because FileCollection has a much nicer ergonomics for plugin users.
- [ ] Other or unforeseen changes necessary

This list is a first draft. If you see other necessary steps or think some of these should be skipped, please comment and lets discuss here.

sergej-koscejev:
Deprecate/delete modelcheck and generate plugins in favor of MpsCheck and MpsGenerate tasks.

Make sure all tasks use `FileCollection` for collections of files (even directories, etc., such as plugin roots) rather than `SetProperty<Directory>` because `FileCollection` has a much nicer ergonomics for plugin users.

sergej-koscejev:
Deprecate the JBR download plugin and related tasks in favor of https://plugins.gradle.org/plugin/com.specificlanguages.jbr-toolchain or some JetBrains-provided solution such as mentioned in #100.

nkoester:
Thanks! I added them to the list.

PR #170, opened 2025-05-23
Use Gradle Version Catalog

nkoester:
This is the first PR in a series of PRs to move the plugin towards a v3, see #169.

Using the Gradle Version Catalog is the basis to migrate parts of PR #161 into the plugin.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ test/generate-build-solution/mps-prj/solutions/my.build/build.xml
 workspace.xml
 **/*_gen*/
 .gradle
+.ideaconfig
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -7,24 +7,18 @@ buildscript {
     }
 
     dependencies {
-        classpath("de.itemis.mps.gradle:git-based-versioning")
+        classpath(libs.itemis.gradle.git.based.versioning)
     }
 }
-
-val kotlinApiVersion by extra { "1.7" }
-val kotlinVersion by extra { "$kotlinApiVersion.10" }
-
 
 plugins {
     groovy
     `java-gradle-plugin`
     `kotlin-dsl`
     `maven-publish`
-    kotlin("jvm") version "1.7.10"
-    id("org.jetbrains.kotlinx.binary-compatibility-validator") version "0.13.2"
+    kotlin("jvm") version libs.versions.kotlin
+    alias(libs.plugins.kotlin.compatibility.validator)
 }
-
-val baseVersion = "1.29.2"
 
 group = "de.itemis.mps"
 
@@ -35,14 +29,14 @@ version = if (!project.hasProperty("useSnapshot") &&
 ) {
     val prefix = when (currentBranch) {
         null, "", "v1.x", "HEAD", "master", "main" -> ""
-        else -> "$currentBranch."
+        else -> "${libs.versions.baseVersion.get()}."
     }
 
     val suffix = ".${GitBasedVersioning.getGitCommitCount()}.${GitBasedVersioning.getGitShortCommitHash()}"
 
-    prefix + baseVersion + suffix
+    prefix + libs.versions.baseVersion.get() + suffix
 } else {
-    "$baseVersion-SNAPSHOT"
+    "${libs.versions.baseVersion.get()}-SNAPSHOT"
 }
 
 val mpsConfiguration = configurations.create("mps")
@@ -58,11 +52,11 @@ dependencyLocking {
 }
 
 dependencies {
-    api("de.itemis.mps.gradle:git-based-versioning")
-    implementation(kotlin("stdlib", version = kotlinVersion))
-    implementation("net.swiftzer.semver:semver:1.1.2")
-    implementation("de.itemis.mps.build-backends:launcher:2.4.0.+")
-    testImplementation("junit:junit:4.13.2")
+    api(libs.itemis.gradle.git.based.versioning)
+    implementation(kotlin("stdlib", version = libs.versions.kotlin.get()))
+    implementation(libs.swiftzer.semver)
+    implementation(libs.itemis.gradle.build.backends.launcher)
+    testImplementation(libs.junit)
 }
 
 tasks.test {
@@ -151,8 +145,8 @@ java {
 }
 
 tasks.withType<KotlinCompile> {
-    kotlinOptions.jvmTarget = "1.8"
-    kotlinOptions.apiVersion = kotlinApiVersion
+    kotlinOptions.jvmTarget = libs.versions.kotlinJVMtarget.get()
+    kotlinOptions.apiVersion = libs.versions.kotlinApi.get()
     kotlinOptions.allWarningsAsErrors = true
 }
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -7,7 +7,7 @@ buildscript {
     }
 
     dependencies {
-        classpath(libs.itemis.gradle.git.based.versioning)
+        classpath("de.itemis.mps.gradle:git-based-versioning")
     }
 }
 
@@ -20,6 +20,8 @@ plugins {
     alias(libs.plugins.kotlin.compatibility.validator)
 }
 
+val baseVersion = "1.29.1"
+
 group = "de.itemis.mps"
 
 val currentBranch : String? = GitBasedVersioning.getGitBranch()
@@ -29,14 +31,14 @@ version = if (!project.hasProperty("useSnapshot") &&
 ) {
     val prefix = when (currentBranch) {
         null, "", "v1.x", "HEAD", "master", "main" -> ""
-        else -> "${libs.versions.baseVersion.get()}."
+        else -> "$currentBranch."
     }
 
     val suffix = ".${GitBasedVersioning.getGitCommitCount()}.${GitBasedVersioning.getGitShortCommitHash()}"
 
-    prefix + libs.versions.baseVersion.get() + suffix
+    prefix + currentBranch + suffix
 } else {
-    "${libs.versions.baseVersion.get()}-SNAPSHOT"
+    "$currentBranch-SNAPSHOT"
 }
 
 val mpsConfiguration = configurations.create("mps")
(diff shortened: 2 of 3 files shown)


build.gradle.kts
```diff
@@ -35,14 +29,14 @@ version = if (!project.hasProperty("useSnapshot") &&
 ) {
     val prefix = when (currentBranch) {
         null, "", "v1.x", "HEAD", "master", "main" -> ""
-        else -> "$currentBranch."
+        else -> "${libs.versions.baseVersion.get()}."
```
sergej-koscejev: This looks like a mistake to me

gradle/libs.versions.toml
```diff
@@ -0,0 +1,27 @@
+[versions]
+# this is our own version we deploy
+baseVersion =  "1.29.2"
+
+#versions
+kotlinApi = "1.7"
+kotlin = "1.7.10"
+kotlinJVMtarget = "1.8"
```
sergej-koscejev: I think the JVM target does not belong here (rather in gradle.properties or in the build script), but in any case its spelling should be fixed:

```suggestion
kotlinJvmTarget = "1.8"
```

build.gradle.kts
```diff
@@ -7,25 +7,19 @@ buildscript {
     }
 
     dependencies {
-        classpath("de.itemis.mps.gradle:git-based-versioning")
+        classpath(libs.itemis.gradle.git.based.versioning)
```
sergej-koscejev: Not sure it makes sense to move this to libs.versions.toml, this is a plugin that's part of the same Git repository, not an external library, so it will never have a version.

gradle/libs.versions.toml
```diff
@@ -0,0 +1,27 @@
...
+
+# libraries
+semver = "1.1.2"
+junit = "4.13.2"
+
+git-based-versioning = "*"
+buildBackends = "2.4.0+"
+
+[plugins]
+kotlin-compatibility-validator = { id = "org.jetbrains.kotlinx.binary-compatibility-validator", version.ref = "kotlinCompatibilityValidator" }
```
sergej-koscejev, line 17: I'd argue that extracting versions that are used only once into `[versions]` brings no real benefits, you can put the version directly here
```suggestion
kotlin-compatibility-validator = "org.jetbrains.kotlinx.binary-compatibility-validator:0.13.2"
```

gradle/libs.versions.toml
```diff
@@ -0,0 +1,27 @@
...
+
+# plugins
+kotlinCompatibilityValidator = "0.13.2"
+
+# libraries
+semver = "1.1.2"
+junit = "4.13.2"
+
+git-based-versioning = "*"
+buildBackends = "2.4.0+"
```
sergej-koscejev: Some names in this file use dashes, some use camel case, this should be made consistent.

gradle/libs.versions.toml
nkoester: I think there are several benefits to also pull single mentioned versions into the TOML, taken from the Gradle docs:

- Type-Safe Accessors
- Centralized Version Management
- Version Separation

But most importantly machine readable versions for renovate or other bots are a core benefit. I think this adds value and would strongly vote to keep it.

gradle/libs.versions.toml
nkoester: The JVM target is also a version. Just not to pull this information apart I would vote to keep it in the TOML. I will adjust the name however.

build.gradle.kts
nkoester: `baseVersion` only provides a `org.gradle.api.provider.Provider` so we have to `get` the version. Is there something not working with this code? It build successfully for me.

gradle/libs.versions.toml
sergej-koscejev: Renovate can read embedded versions, and we don’t need to access versions in a type-safe manner. I don’t understand the benefit of version separation here either. To the contrary, things that belong together should stay together.

gradle/libs.versions.toml
sergej-koscejev: Also, the correct syntax is 2.4.0.+, Gradle is confusing here.
```suggestion
buildBackends = "2.4.0.+"
```

gradle/libs.versions.toml
sergej-koscejev: Ok

gradle/libs.versions.toml
nkoester: I guess it depends whether or not we agree on using the version catalog? I am fine in dropping this PR ifthere is general doubt from your perspective...

But if we want to use a version catalog then leaving this single version out would be the actual separation from my perspective.

gradle/libs.versions.toml
nkoester: Good catch! I accidentally dropped the dot in my change.

gradle/libs.versions.toml
sergej-koscejev: I don't mean this single version should be left out but all versions that are only used from one place (e.g. semver, too).

In my view, extracting the versions into [versions] section makes sense for versions of related libraries built from one repo, such as Kotlin or JUnit 5 (which has `org.junit.jupiter:junit-jupiter-api`, `org.junit.jupiter:junit-jupiter-engine`, `org.junit.jupiter:junit-jupiter-params` etc. all using the same version).

I agree that it makes sense to extract all dependencies into the version catalog to avoid projects accidentally using different and slightly incompatible versions. But I don't think it makes sense to extract _all_ versions into the [versions] section, only where it helps avoid duplication.

build.gradle.kts
sergej-koscejev: No, I mean the code used to put the current branch here before and add the base version two lines down, now it adds the base version twice.

gradle/libs.versions.toml
sergej-koscejev: Also, I wouldn't expect to find _our own_ version in libs.versions.toml, that's after all for libraries, i.e. dependencies.

build.gradle.kts
nkoester: You are right, the version catalog is intended for dependency versions and not our own version. I will revert this part.

gradle/libs.versions.toml
nkoester: After async discussion we concluded that we will add all versions to the catalog but leave out the own version.

build.gradle.kts
```diff
@@ -11,20 +11,16 @@ buildscript {
...
     `kotlin-dsl`
     `maven-publish`
-    kotlin("jvm") version "1.7.10"
-    id("org.jetbrains.kotlinx.binary-compatibility-validator") version "0.13.2"
+    kotlin("jvm") version libs.versions.kotlin
+    alias(libs.plugins.kotlin.compatibility.validator)
 }
 
-val baseVersion = "1.29.2"
+val baseVersion = "1.29.1"
```
sergej-koscejev:
```suggestion
val baseVersion = "1.29.2"
```

build.gradle.kts
```diff
@@ -40,9 +36,9 @@ version = if (!project.hasProperty("useSnapshot") &&
 
     val suffix = ".${GitBasedVersioning.getGitCommitCount()}.${GitBasedVersioning.getGitShortCommitHash()}"
 
-    prefix + baseVersion + suffix
+    prefix + currentBranch + suffix
```
sergej-koscejev: Is this change intended?

build.gradle.kts
```diff
@@ -40,9 +36,9 @@ version = if (!project.hasProperty("useSnapshot") &&
...
 } else {
-    "$baseVersion-SNAPSHOT"
+    "$currentBranch-SNAPSHOT"
```
sergej-koscejev: Is this change intended?

sergej-koscejev (changes requested):

.gitignore
```diff
@@ -18,3 +18,4 @@ test/generate-build-solution/mps-prj/solutions/my.build/build.xml
 workspace.xml
 **/*_gen*/
 .gradle
+.ideaconfig
```
sergej-koscejev, line 21: Out of curiosity, what is this file? I don't think I ever saw it anywhere.

nkoester: I have multiple IntelliJ Idea config prefixes (often one per project) to isolate them. I keep them within the checkouts in the `.ideaconfig` folder. I also do the same for MPS projects with `.mpsconfig`.

build.gradle.kts
nkoester: I am sorry - I was not attentive and replaced the wrong variables. Will fix it all.

build.gradle.kts
```diff
@@ -35,7 +31,7 @@ version = if (!project.hasProperty("useSnapshot") &&
 ) {
     val prefix = when (currentBranch) {
         null, "", "v1.x", "HEAD", "master", "main" -> ""
-        else -> "$currentBranch."
+        else -> "$baseVersion."
```
sergej-koscejev: I think this is still unintended? Otherwise the base version will be duplicated for code built on branches.

build.gradle.kts
```diff
@@ -58,11 +54,11 @@ dependencyLocking {
 }
 
 dependencies {
-    api("de.itemis.mps.gradle:git-based-versioning")
-    implementation(kotlin("stdlib", version = kotlinVersion))
-    implementation("net.swiftzer.semver:semver:1.1.2")
-    implementation("de.itemis.mps.build-backends:launcher:2.4.0.+")
-    testImplementation("junit:junit:4.13.2")
+    api(libs.itemis.gradle.git.based.versioning)
+    implementation(kotlin("stdlib", version = libs.versions.kotlin.get()))
```
sergej-koscejev: Should we move kotlin stdlib to libs.versions.toml as well?

nkoester: I was not sure what extra things are happening in the `kotlin()` call so I left it as is. Thanks for the pointer! 

Before I change it - I assume these two calls are identical, correct?

```
implementation(kotlin("stdlib", version = "1.7.10"))
```

VS

```
implementation("org.jetbrains.kotlin:kotlin-stdlib:1.7.10")
```

If yes, I can know how to move it to the TOML.

sergej-koscejev: Yes, they should be equivalent.

sergej-koscejev (approved):

PR #171, opened 2025-07-20
Upgrade to Gradle 8.14.3

wjglerum:
Bump Gradle to the latest version, see https://docs.gradle.org/8.14.3/release-notes.html

And also fixes some deprecations while we are at it:
- Removes deprecated usage of `Task.getProject()`
- Removes deprecated usage of `Project.exec()`
- Removes deprecated Kotlin compiler options

Started this cause I was trying out the gradle 9 release candidates and some things are deprecated/removed.


sergej-koscejev:
The commit message says '8.14.2' but upgrades to 8.14.3.

According to Gradle docs, ExecOperations has been available since Gradle 6.0 so you should be free to use it, the plugin will only run on something like 7.x or newer anyway (or maybe even 8.x), as far as I know.

wjglerum:
> The commit message says '8.14.2' but upgrades to 8.14.3.

whoops will fix that 😄 

> According to Gradle docs, ExecOperations has been available since Gradle 6.0 so you should be free to use it, the plugin will only run on something like 7.x or newer anyway (or maybe even 8.x), as far as I know.

Alright will try to get rid of the deprecations too then 👍

wjglerum:
> According to Gradle docs, ExecOperations has been available since Gradle 6.0 so you should be free to use it, the plugin will only run on something like 7.x or newer anyway (or maybe even 8.x), as far as I know.

@sergej-koscejev done, leveraged property injection to inject the `ExecOperations` where needed, see https://docs.gradle.org/current/userguide/service_injection.html#property_injection

src/main/kotlin/de/itemis/mps/gradle/RunAntScript.kt
```diff
@@ -70,30 +71,30 @@ open class RunAntScript : DefaultTask() {
 
         val targets = if (incremental) { targets - "clean" } else { targets }
 
-        project.runAnt(executable, project.rootDir, allArgs + "-buildfile" + project.file(script).toString() + targets,
+        project.runAnt(execOperations, executable, project.rootDir, allArgs + "-buildfile" + project.file(script).toString() + targets,
```
sergej-koscejev, line 74: We shouldn't need Project here anymore, it was only used to call javaexec.

src/main/kotlin/de/itemis/mps/gradle/downloadJBR/Plugin.kt
```diff
@@ -21,7 +22,11 @@ open class DownloadJbrConfiguration @Inject constructor(objects: ObjectFactory)
         }
 }
 
-open class DownloadJbrProjectPlugin : Plugin<Project> {
+abstract class DownloadJbrProjectPlugin : Plugin<Project> {
+
+    @get:Inject
+    abstract val execOperations: ExecOperations
```
sergej-koscejev:
```suggestion
    protected abstract val execOperations: ExecOperations
```

src/main/kotlin/de/itemis/mps/gradle/RunAntScript.kt
```diff
@@ -28,6 +27,8 @@ open class RunAntScript : DefaultTask() {
     var includeDefaultClasspath = true
     @Optional @Input
     var executable: Any? = null
+    @get:Inject
+    abstract val execOperations: ExecOperations
```
sergej-koscejev:
```suggestion
    protected abstract val execOperations: ExecOperations
```

src/main/kotlin/de/itemis/mps/gradle/runmigrations/Plugin.kt
```diff
@@ -205,7 +209,7 @@ open class RunMigrationsMpsProjectPlugin : Plugin<Project> {
                         }
 
                         runAnt(
-                            extension.javaExec, temporaryDir, args = listOf(),
+                            execOperations,extension.javaExec, temporaryDir, args = listOf(),
```
sergej-koscejev:
```suggestion
                            execOperations, extension.javaExec, temporaryDir, args = listOf(),
```

src/main/kotlin/de/itemis/mps/gradle/RunAntScript.kt
wjglerum: That's also what I thought, but we actually do use it to get some properties from the project in the internal method

sergej-koscejev:
Looks good, please add a (short) changelog entry and clean up the commit history.

sergej-koscejev (approved):

wjglerum:
> Looks good, please add a (short) changelog entry and clean up the commit history.

Done! ✅

sergej-koscejev (approved):

PR #178, opened 2026-04-23
TeamCity: port build configs to in-repo Kotlin DSL

sergej-koscejev:
## Summary

- Moves the `mps-gradle-plugin` and `git-based-versioning` TeamCity build configurations from the TC UI into `.teamcity/settings.kts` (Kotlin DSL).
- Keeps full parity with the current v1.x behavior: the plugin build runs `setTeamCityBuildNumber build` and then `publish`; `git-based-versioning` runs `:git-based-versioning:build :git-based-versioning:publish`.
- Pom layout mirrors `mbeddr/publish-mps-prereleases/.teamcity/pom.xml` — parent `org.jetbrains.teamcity:configs-dsl-kotlin-parent:1.0-SNAPSHOT` via `https://mps.builds.itemis.cloud/app/dsl-plugins-repository`.
- Pairs with the master PR (same change, without the `Publish` step) so master can later diverge for #177 (selective on-demand publishing of included builds) without touching v1.x.

## Handover checklist for TeamCity

These steps apply once both this PR and the master PR are merged:

1. On the newly-created `Mbeddr_Tooling_MpsGradlePlugin` subproject in TeamCity, switch Versioned Settings `synchronizationMode` from `useParentProjectSettings` to **Synchronization enabled**. Point at the existing VCS root (`Mbeddr_Tooling_MpsGradlePlugin`, which targets `https://github.com/mbeddr/mps-gradle-plugin`). Turn on **"Use settings from the branch being built"**.
2. TC imports the DSL; two new build configs appear. Trigger each manually on v1.x and master and compare against the last legacy runs (build log, `##teamcity[buildNumber '…']`, artifact list, GitHub commit status).
3. After parity, archive or delete the two legacy configs (`Mbeddr2_Mbeddr_Gradle_MpsGradlePlugin`, `Mbeddr2_Mbeddr_Gradle_GitBasedVersioning`).

Note: the DSL reuses the inherited `%system.github.token%` for the commit-status publisher. Worth confirming this is the same token the current legacy config's per-build-config `secure:github_access_token` holds; if not, we'll need a project-scoped credentials entry before archiving the old config.

## Test plan

- [ ] `cd .teamcity && mvn teamcity-configs:generate` produces XML under `target/generated-configs/` matching the live configs' semantics.
- [ ] After TC sync on v1.x, plugin build runs Build + Publish, publishes a `maven-mps-snapshots` artifact, GitHub commit status updates.
[...]

PR #177, opened 2026-04-23
3.0.0: task-type interfaces, Gradle 9, drop legacy plugins

sergej-koscejev:
## Summary

Targets 3.0.0. Grouped:

### New task-type interfaces

- `MpsTask` (`mpsHome`, `mpsVersion`, `javaLauncher`, `pluginRoots`,
  `folderMacros`, `logLevel`) and `MpsProjectTask` (extends `MpsTask`, adds
  `projectLocation`) implemented by all MPS task types including
  `RunAntScript`. Both interfaces support `tasks.withType` bulk configuration.
- Published as a separate `de.itemis.mps:mps-gradle-plugin-api:1.0.0` artifact
  (new `:mps-gradle-plugin-api` included build).
- The `de.itemis.mps.gradle.common` plugin sets the `MpsTask.logLevel`
  convention to `gradle.startParameter.logLevel`.

### Platform bump

- Requires Gradle 9 and Java 17 (previously Gradle 8 and Java 11).

### `RunAntScript`

- `script`, `targets`, `scriptArgs`, `scriptClasspath` converted to lazy
  Gradle properties. `script` is a `RegularFileProperty` annotated `@InputFile`,
  so wiring `script.set(otherTask.flatMap { ... })` establishes the task
  dependency automatically.
- Added `workingDirectory: DirectoryProperty` (default: root project directory).
- Now requires `mpsHome`; the Ant classpath defaults to `lib/ant/lib/*.jar`
  and `lib/*.jar` under `mpsHome`. `mps.home` and `mps_home` are passed as
  Ant properties.
- Uses inherited `javaLauncher` and `logLevel`; `-Dmps.ant.log` is derived
  from `logLevel`.
- Removed `executable`, `incremental`, `includeDefaultArgs`,
  `includeDefaultClasspath`, and the `itemis.mps.gradle.ant.defaultScriptArgs`
  / `defaultScriptClasspath` / `defaultJavaExecutable` extended properties.
  Use `tasks.withType<RunAntScript>().configureEach { ... }` to share
  configuration.

### Removed

- Old plugins `generate-models`, `modelcheck`, `run-migrations` — use the
  task types (`MpsGenerate`, `MpsCheck`, `MpsMigrate`, `Remigrate`) directly.
- `download-jbr` plugin, `DownloadJbrForPlatform` task, and the `downloadJBR`
  package — use the
[...]

PR #179, opened 2026-04-23
TeamCity: port build configs to in-repo Kotlin DSL

sergej-koscejev:
## Summary

- Moves the `mps-gradle-plugin` and `git-based-versioning` TeamCity build configurations from the TC UI into `.teamcity/settings.kts` (Kotlin DSL).
- Master's plugin build only runs the `Build` step — no `Publish`. That matches today's UI-managed behavior (the legacy config has a step condition that skips publishing on master) and sets up #177 to land the selective on-demand publish as a follow-up without rebuilding the whole DSL.
- `git-based-versioning` build is identical to v1.x's.
- Pairs with the v1.x PR #178 (same structure, with the v1.x-specific `Publish` step added).

## Handover checklist for TeamCity

See the v1.x PR for the full checklist. Both PRs need to be merged before enabling versioned settings on `Mbeddr_Tooling_MpsGradlePlugin` (TC loads the default branch — master — to determine the set of build configs, then per-branch DSL for runtime behavior).

## Test plan

- [ ] `cd .teamcity && mvn teamcity-configs:generate` produces XML under `target/generated-configs/`; plugin build has only the Build step, `git-based-versioning` build matches the current legacy config.
- [ ] After TC sync on master, plugin build runs Build only, no artifacts land in `maven-mps-snapshots`, GitHub commit status updates.
- [ ] After TC sync on master, `git-based-versioning` build runs successfully when triggered manually.

## Follow-ups

- #177 will wire selective on-demand publishing of included builds on master.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nkoester (approved):
I've reviewed the changes in this PR. Sergej and I had a call on March 23rd to discuss the context, and we've covered all the $3.0$-related updates. Thanks for your work! LGTM!

